PR #106, opened 2017-07-18
Timeout Adjustment algorithm

oudod:


DarkEld3r:
Some changes to the timeout adjusters: https://github.com/exonum/exonum/pull/256.

PR #108, opened 2017-07-18
proofreading

Maria-Nosyk:
Fixed some grammar, checked readability of the text in whole, conducted semantic analyses of the ambiguous parts


src/advanced/node-management.md
```diff
@@ -295,7 +295,7 @@ Response is a JSON object with the following fields:
 }
 ```
 
-### Private endpoints
+## Private endpoints
```
oudod: Please make header levels the same (for Private endpoints and Public endpoints). IMO level 2 is better.

PR #109, opened 2017-07-18
proofreading of node-management

Maria-Nosyk:
Checked grammar, clearness of language and conducted semantic analyses of ambiguous parts.


slowli:
What was wrong with #108? Please don't create multiple requests for the same issue.

PR #126, opened 2017-09-26
Describe serialization of structure composition

oudod:


src/architecture/serialization.md
```diff
@@ -243,6 +243,33 @@ in the correspondence with [the validation rules](#segment-validation-rules).
...
+encoding_struct! {
+  struct Pair {
+    const SIZE = 8;
+    field first: u32 [00 => 04]
+    field second: u32 [04 => 08]
+  }
+}
+
+encoding_struct! {
+  struct Pairs{
```
slowli: Nitpick: Missing space before `{`.

src/architecture/serialization.md
```diff
@@ -243,6 +243,33 @@ in the correspondence with [the validation rules](#segment-validation-rules).
...
+    const SIZE = 8;
+    field inner : Vec<Pair> [00 => 08]
```
slowli: Nitpick: Obsolete space before `:`.

src/architecture/serialization.md
```diff
@@ -243,6 +243,33 @@ in the correspondence with [the validation rules](#segment-validation-rules).
...
+  }
+}
+```
+
+Here `Pair` is used as a base type in `Pairs` declaration.
+
+!!! note
+    `message!` macro may be used instead of `encoding_struct!` for the outer
+    structure.
```
slowli: Do you mean this is the case when the outer "structure" is actually a message? (That is, a message can incorporate structure fields in its body.) If this is the case, maybe describe it like:

> Analogously, fields derived from structures defined with `encoding_struct!` macro can be used within messages.

I'd also add an explicit example just to be clear.

src/architecture/serialization.md
```diff
@@ -243,6 +243,33 @@ in the correspondence with [the validation rules](#segment-validation-rules).
     binary format all fields of such structure are placed in the header, its body
     is empty. So such a structure is fixed-length.
 
+#### Structure composition
+
+`encoding_struct!` macro allows to make a composition of existing structures:
```
slowli: Nitpick: "...allows to compose existing structures" sounds a bit more natural to me.

PR #127, opened 2017-09-27
Update tutorial to make Wallet immutable

slowli:
...according to exonum/cryptocurrency#35.

### DIFF
--- a/src/get-started/create-service.md
+++ b/src/get-started/create-service.md
@@ -239,7 +239,7 @@ impl Wallet {
 ```
 
 We have added two methods: one to increase the wallet balance and another one
-to decrease it. These methods are *immutable*; they consume an old instance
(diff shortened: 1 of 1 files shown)


src/get-started/create-service.md
```diff
@@ -226,22 +226,21 @@ so we add methods to the `Wallet` type:
...
+        Self::new(self.pub_key(), self.name(), balance)
     }
 }
 ```
 
 We have added two methods: one to increase the wallet balance and another one
-to decrease it. We used `Field::write` because the data in the structs processed
-by `encoding_struct` is stored as a binary blob and we need to overwrite it
-in-place.
+to decrease it. These methods are *immutable*; they consume an old instance
```
Maria-Nosyk: an old instance --> the old...

PR #128, opened 2017-10-10
Added rocksdb support into installation guide

aleksuss:


src/get-started/install.md
```diff
@@ -84,6 +85,16 @@ You may also run the extended test suite located in the `sandbox` directory:
 cargo test --manifest-path sandbox/Cargo.toml
 ```
 
+Since version `0.2.0` Exonum supports RocksDB as alternative data storage.
+To enable RocksDB support you need to pass  additional parameter to cargo:
```
Maria-Nosyk: "...as alternative data storage. To enable RocksDB support..." --> "...as an alternative data storage, to enable RocksDB support..."

slowli: Are you sure? Seems like these are 2 separate sentences.

src/get-started/install.md
```diff
@@ -84,6 +85,16 @@ You may also run the extended test suite located in the `sandbox` directory:
 cargo test --manifest-path sandbox/Cargo.toml
 ```
 
+Since version `0.2.0` Exonum supports RocksDB as alternative data storage.
```
slowli: Could you please add a link to the release on GitHub or in Cargo here?

src/get-started/install.md
```diff
@@ -19,6 +19,7 @@ to these projects, testing them, and developing using Exonum.
 Exonum depends on the following third-party system libraries:
 
 - [LevelDB][leveldb] (persistent storage)
+- [RocksDB][rocksdb] (persistent storage)
```
slowli: Could you please clarify: is LevelDB and RocksDB optional or necessary dependencies? It seems that both are optional judging by [Cargo.toml](https://github.com/exonum/exonum/blob/master/exonum/Cargo.toml) in the `exonum` crate.

src/get-started/install.md
Maria-Nosyk: @slowli ok, let it be 
"+As of version `0.2.0` Exonum supports RocksDB as alternative data storage.
+To enable RocksDB support you need to pass  additional parameter to cargo:"

src/get-started/install.md
aleksuss: @slowli `LevelDB` is default feature https://github.com/exonum/exonum/blob/master/exonum/Cargo.toml#L54

src/get-started/install.md
```diff
@@ -84,6 +85,22 @@ You may also run the extended test suite located in the `sandbox` directory:
 cargo test --manifest-path sandbox/Cargo.toml
 ```
 
+Since version [0.2.0][rel0.2.0] Exonum supports RocksDB as alternative data storage.
```
Maria-Nosyk: Since --> As of

src/get-started/install.md
```diff
@@ -84,6 +85,22 @@ You may also run the extended test suite located in the `sandbox` directory:
...
+```shell
+cargo test --manifest-path exonum/Cargo.toml --features rocksdb
+```
+and for extended test suite:
+```shell
+cargo test --manifest-path sandbox/Cargo.toml --features rocksdb
+```
+
+If you want to use Exonum framework with RocksDB support as dependency in you project 
+you should add into `Cargo.toml` line like this:
```
Maria-Nosyk: "...RocksDB support as dependency in you project you should add into `Cargo.toml` line like this:" -->
"RocksDB support as a dependency in your project, you should add the following line into `Cargo.toml`:"

src/get-started/install.md
slowli: Thanks! But it's still possible to compile and use Exonum without it, right? If so, I'd advise adding something like this below the list of dependencies:

> `LevelDB` and `RocksDB` are used as alternative storage engines, with `LevelDB` used by default. You need to install only the storage engine(s) you are intending to use.

aleksuss: Yes, you are right.

PR #129, opened 2017-10-11
add mention of RocksDB to documentation

Maria-Nosyk:


src/architecture/services.md
```diff
@@ -131,19 +148,22 @@ cryptocurrency service persists account balances, which are changed by transfer
 and issuance transactions.
 
 Exonum persists blockchain state in a global key-value storage implemented
-using [LevelDB][leveldb]. Each service needs to define a set of data collections
+using [LevelDB][leveldb] and [RocksDB][rocksdb]. Each service needs to define a
```
slowli: Nitpick: This may sound like *both* LevelDB and RocksDB are needed for the KVS implementation, which is not true. Maybe change to "...storage implemented with one of supported backends (LevelDB and RocksDB as of Exonum 0.2)"?

src/architecture/storage.md
```diff
@@ -232,7 +239,7 @@ and suffixes are used to distinguish tables within the group.
 !!! note "Example"
     Key `key` at the table named `BTC` (`0x42 0x54 0x43` in ASCII) at
     the table group `0x03` for the service with ID `0x00 0x01` matches the
-    following key in the LevelDB map:
+    following key in the LevelDB map (and currently also in the RocksDB map):
```
slowli: Is the logic supposed to change in the future releases? If there is a finalized roadmap how, I'd appreciate if you put it in another note below.

PR #130, opened 2017-10-19
Rename service initialization handler

slowli:
...from `handle_genesis_block()` to `initialize()` according to changes in Exonum core.

### DIFF
--- a/src/architecture/services.md
+++ b/src/architecture/services.md
@@ -420,7 +420,7 @@ the blockchain initialization. A result of the method call for each service
 is recorded under [the string service identifier](#service-identifiers)
 in the configuration. The resulting initial configuration is augmented
 by non-service parameters (such as public keys of the validators) and is
-recorded in the genesis block, hence the method name.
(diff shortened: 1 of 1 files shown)


src/architecture/services.md
```diff
@@ -402,18 +402,17 @@ The service, which `tx_from_raw` method
...
 
-fn handle_genesis_block(&self, fork: &mut Fork)
-                        -> Value {
+fn initialize(&self, fork: &mut Fork) -> Value {
     Value::Null
 }
 ```
 
-`handle_genesis_block` returns an initial
+`initialize` returns an initial
```
dmitry-timofeev: Can't add inline, but "hence the method name" no longer makes sense in https://github.com/exonum/exonum-doc/pull/130/files#diff-0dff3c1379cd0e3496d9af1f140a731bR423

slowli: Jeez, totally missed that. Thanks for noticing!

PR #132, opened 2017-10-24
Clarify parameters for the blocks-in-range request

quicky84:
Imagine, you have a chain with empty first 1001 blocks, you want to retrieve all non-empty blocks. You pull blocks starting from 0 with increment of 1000. There is no way to say whether you have exhausted all blocks in the chain or you haven't reached non-empty blocks yet. So should you keep pulling or stop? Thus one needs to start pulling from the end until 0 is reached. It is unclear from the documentation that skipping the `latest` param will result in a pull from the very end of the chain.

### DIFF
--- a/src/advanced/node-management.md
+++ b/src/advanced/node-management.md
@@ -544,11 +544,11 @@ Returns the headers of the consecutive `count` blocks up to the `latest` height.
   The number of blocks to return. Should not be greater than
   [`MAX_BLOCKS_PER_REQUEST`][github_explorer]
 - **skip_empty_blocks**: bool=  
-  If `true`, then only non-empty blocks are returned. Default value is `false`
+  If `true`, then only non-empty blocks are returned. The default value is `false`
 - **latest**: integer=  
   The maximum height of the returned blocks. The blocks are returned
   in reverse order, starting from the `latest` and at least up to the `latest -
-  count + 1`. Default value is the height of the latest block in the blockchain
+  count + 1`. The default value is the height of the latest block in the blockchain
 
 #### Response
 
--- a/src/advanced/node-management.md
+++ b/src/advanced/node-management.md
@@ -543,12 +543,12 @@ Returns the headers of the consecutive `count` blocks up to the `latest` height.
 - **count**: integer  
   The number of blocks to return. Should not be greater than
   [`MAX_BLOCKS_PER_REQUEST`][github_explorer]
-- **skip_empty_blocks**: bool  
-  If `true`, then only non-empty blocks are returned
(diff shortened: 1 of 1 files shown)


quicky84:
By the way the problem persists even if you start pulling blocks from the tail of the chain. Say, you have 2000 blocks with 1001 last blocks empty (but it's only known server-side) and you specify `count=1000` and `skip_empty_blocks=true`. The response will be `[]`, on the client-side it's impossible to conclude whether the chain is exhausted and thus contains only empty blocks or you just pulled from a range of empty blocks.

The only viable option (right now) to pull all blocks from the chain is to start pulling from the tail and then filter on `tx_count`. This actually defeats the purpose of `skip_empy_blocks`.

slowli:
Could you please open a corresponding issue in the [`exonum`](https://github.com/exonum/exonum/) repository? I'd say a good solution would involve implementing common practices in retrieving ranges, with the response looking something like:
```js
{
  "blocks": [ /* blocks here */ ],
  "from_index": 1000,
  "to_index": 2000,
  "skip_empty_blocks": true
}
```

src/advanced/node-management.md
```diff
@@ -543,12 +543,12 @@ Returns the headers of the consecutive `count` blocks up to the `latest` height.
 - **count**: integer  
   The number of blocks to return. Should not be greater than
   [`MAX_BLOCKS_PER_REQUEST`][github_explorer]
-- **skip_empty_blocks**: bool  
-  If `true`, then only non-empty blocks are returned
-- **latest**: integer  
+- **skip_empty_blocks**: bool=  
+  If `true`, then only non-empty blocks are returned. Default value is `false`
```
slowli: *The* default value?

quicky84: Uhm, I am not a native speaker, always have troubles with (the? : ) articles

src/advanced/node-management.md
```diff
@@ -543,12 +543,12 @@ Returns the headers of the consecutive `count` blocks up to the `latest` height.
...
+- **latest**: integer=  
   The maximum height of the returned blocks. The blocks are returned
   in reverse order, starting from the `latest` and at least up to the `latest -
-  count + 1`
+  count + 1`. Default value is the height of the latest block in the blockchain
```
Maria-Nosyk: same here

slowli:
@quicky84, do you mind if I add `the` articles directly to your branch in order to merge this PR? Or you may add them yourself :smiley:

quicky84:
Please do. I appreciate this.

PR #136, opened 2017-11-11
Update info on persistence according to 0.3 changes

slowli:
This PR updates information in docs about the persistence engine used by Exonum. In particular, LevelDB support is mentioned as removed.

### DIFF
--- a/src/architecture/storage.md
+++ b/src/architecture/storage.md
@@ -226,20 +226,24 @@ Every table is uniquely identified by a compound identifier, which is used
 to map table keys into a column family and its keys in the underlying
 low-level storage. A table identifier consists of 2 parts:
 
-- String name, which is mapped 1-to-1 to a column family.
+- **String name,** which is mapped 1-to-1 to a column family.
   The name may contain uppercase and lowercase Latin letters, digits,
-  and underscores `_`.
-- Optional prefix presented as a sequence of bytes (`Vec<u8>` in Rust terms).
+  underscores `_`, and periods `.`. By convention, table names in services should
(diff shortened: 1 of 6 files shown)


src/architecture/storage.md
```diff
@@ -200,63 +200,54 @@ Exonum uses third-party database engines to persist blockchain state
...
+to map table keys into a column family and its keys in the underlying
+low-level storage. A table identifier consists of 2 parts:
 
-The table prefix consists of [the service ID](services.md#service-identifiers)
-and an internal identifier inside the
-service.
-All tables created with the same prefix will be the views of the same data.
+- String name, which is mapped 1-to-1 to a column family.
+  The name may contain uppercase and lowercase Latin letters, digits,
+  and underscores `_`.
```
slowli: This is where it gets a bit unclear. [The source](https://github.com/exonum/exonum/blob/v0.3/exonum/src/storage/base_index.rs#L249) shows that periods `.` are supported in the table names, too; and, in fact, is used in [all "core" tables](https://github.com/exonum/exonum/blob/v0.3/exonum/src/blockchain/schema.rs) and, say, [tables of the configuration service](https://github.com/exonum/exonum-configuration/blob/v0.3/src/lib.rs) (but not in code samples). Should every table name prefixed with `<service_name>.` as a convention?

DarkEld3r: > Should every table name prefixed with <service_name>. as a convention?

Yes, it should. Ideally we want to append a service prefix automatically, but such improvement will not be available soon I guess.

PR #139, opened 2017-12-18
Update the `Block in Range` section

quicky84:
A documentation update regarding changes in the behavior of the public endpoint, see  exonum/exonum#351.


slowli:
@DarkEld3r Could you please take a look whether this description corresponds to what was released in v0.6? It seems so, in which case this PR can finally be merged.

PR #160, opened 2018-01-21
Remove unnecessary "!!! note" mkdocs tag

lazaridiscom:
The exonum documentation looks & reads nice, but it uses far too much formatting. A clear example is this "note". This special-format sections should be kept minimal, reserved for really critical information (e.g.: "Security" and "Privacy").

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -4,9 +4,7 @@ The [Exonum core project][exonum] and [other public Exonum repositories][exonum-
 operate an open contributor model where anyone is welcome to contribute
 towards development in the form of peer review, testing and patches.
 This document explains the practical process and guidelines for contributing.
-
-!!! note
-    This contribution guide is partially derived from [the Bitcoin contribution guide][btc-contrib].
+It is partially derived from [the Bitcoin contribution guide][btc-contrib].
 
 !!! summary "TL;DR"
     Use pull requests, use [git commit message conventions][git:messages],


src/contributing.md
```diff
@@ -4,9 +4,7 @@ The [Exonum core project][exonum] and [other public Exonum repositories][exonum-
 operate an open contributor model where anyone is welcome to contribute
 towards development in the form of peer review, testing and patches.
 This document explains the practical process and guidelines for contributing.
-
-!!! note
-    This contribution guide is partially derived from [the Bitcoin contribution guide][btc-contrib].
+It is partially derived from [the Bitcoin contribution guide][btc-contrib].
```
slowli: I'd use "This contribution guide" or "This guide" instead of "It"; imo, "it" may be a little confusing since paragraphs should make sense in insolation. Alternatively, you may move the sentence to the previous paragraph, although I would advise not to do this: the first para of the page is automatically used as the page description in meta, so bloating it would not be wise.

lazaridiscom: >Alternatively, you may move the sentence to the previous paragraph

It is already in the 1st paragraph, see the formatted document. And its a credit, so it possibly should have good visibility.

lazaridiscom:
@slowli 
> If there are other cases of over-formatting in your opinion, would you mind pointing to them or including fixes into this PR?

I understand your overall elaborations. As to your question, this is possibly a matter of taste. Those admonitions "break" me personally out of my read-flow. In the given "contributions.md", I personally think that all of them are not needed.

Another point is, that if I assess the "Tips" as "trivial", I start to overlook the tip boxes anyway - missing this way possibly a relevant information at some (tip) point.

And finally: the design of the admonitions feels to me a bit "too much". Would like them a little bit smoother. But the again: all this is a matter of taste, too.

src/contributing.md
slowli: Oh sorry, missed that (didn't look at the rendered docs). Yeah, the sentence after your edits looks completely fine.

lazaridiscom: np! Markdown authoring is very nice, but I get myself sometimes confused from the render-discrepancies.

slowli:
I understand your position, but, as you say, this is largely a matter of taste.

>Another point is, that if I assess the "Tips" as "trivial", I start to overlook the tip boxes anyway - missing this way possibly a relevant information at some (tip) point.

Tips are intended to concentrate advice, say, how to use Exonum. If you don't need such tips, imo, it's better if they are located easily as you may skip them, as you say. Conversely, if you look for, say, examples, the docs could benefit if they can be located easily on the page.

Now, there may be a problem that some of tips / examples / warnings / ... are insufficiently logically separated from the surrounding text (so they should be incorporated in it), or that they are unneeded altogether (so they should be removed). The one you've fixed belongs to the first category; one can argue that the "Please refer to the Git manual..." tip on the same page belongs to the second category.

>And finally: the design of the admonitions feels to me a bit "too much". Would like them a little bit smoother. But the again: all this is a matter of taste, too.

The admonitions are taken directly from the [`mkdocs-material`](https://squidfunk.github.io/mkdocs-material/extensions/admonition/) theme. I personally find them quite fitting for the material design.

lazaridiscom:
Agreed that "Please refer to the Git manual..." should be removed.

As for the default mkdocs-formatting (which i personally assess as too "fancy"), I've found a workaround: Firefox reader mode, which strips away any of the mkdocs theme formatting, so I can focus on the text.

PR #162, opened 2018-01-23
update date in copyright

boguslavsky:


slowli:
Could you please update the year in README as well?

boguslavsky:
@slowli done

PR #172, opened 2018-02-09
Rename num_votes to num_validators

ivan-ochc:
Update of documentation according to https://github.com/exonum/exonum-configuration/pull/83

### DIFF
--- a/src/advanced/configuration-updater.md
+++ b/src/advanced/configuration-updater.md
@@ -128,8 +128,8 @@ config][config_propose] serialization. It has the following fields:
   String containing JSON serialization of proposed configuration.
 - **votes_history_hash**: Hash  
   Hash of the proposed configuration.
-- **num_votes**: integer  
-  Number of votes for the proposed configuration.
+- **num_validators**: integer  
+  Number of available validators.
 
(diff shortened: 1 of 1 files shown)


src/advanced/configuration-updater.md
```diff
@@ -128,8 +128,8 @@ config][config_propose] serialization. It has the following fields:
   String containing JSON serialization of proposed configuration.
 - **votes_history_hash**: Hash  
   Hash of the proposed configuration.
-- **num_votes**: integer  
-  Number of votes for the proposed configuration.
+- **num_validators**: integer  
+  Number of available validators.
```
slowli: "Available validators" isn't very descriptive wording... Do I understand correctly that `num_validators` is the number of validators that can vote for the proposal (or, in other words, the maximum number of possible votes for the proposal)? If so, propose to elaborate using these words.

ivan-ochc: Will "Number of validators that can vote for the proposal" be fine?

slowli: Yeah 👍 .

PR #177, opened 2018-02-15
Describe time service

slowli:
This PR describes the [time oracle service](https://github.com/exonum/exonum/tree/master/services/time). The text is an improved version of `TUTORIAL.md` in the repo; the idea is to move the document here.


src/advanced/time.md
```diff
@@ -0,0 +1,284 @@
...
+  replace the consolidated time with `t`.
+
+Thus, the consolidated time can be updated after each transaction with
+the actual time from any validator node. The procedure takes into account
+possible changes in the validators list, ensures monotony of `time`, and
+is tolerant to the malicious behavior of validator nodes.
+
+## Proof of Correctness
+
+Denote `T` the list of current validators’ timestamps sorted in the decreasing order,
```
Maria-Nosyk: "Denote `T` the list of current validators’ timestamps" --> probably the following is meant "Denote `T` is the list of current validators’ timestamps "??

src/advanced/time.md
```diff
@@ -0,0 +1,284 @@
...
+- The time of an honest node, or
+- The time between the timestamps of two honest nodes
+  (and therefore such a time can be considered reliable).
+
+For practical reasons, we always choose the timestamp with index `f + 1`,
+since this value is reliable and at the same time the most recent one.
+
+## Discussion
+
+The validator nodes can potentially generate and send a transaction to update
```
Maria-Nosyk: "send a transaction" --> "send transactions"
"the nodes send the transaction" --> "the nodes send the transactions"

src/advanced/time.md
```diff
@@ -0,0 +1,287 @@
...
+- Time updates are tied to creation of `Precommit` messages. In the case
+  of a large delay in block acceptance, all the transactions therein will
+  be executed with the same outdated time value.
+
+### Time Oracle Service
+
+Each validator at a specific time sends a transaction indicating its local time
+(usually immediately after the commit of each block). Exonum storage contains
+an index with the most current time indicated separately by each validator.
+A median of these index values is stored separately; it is considered the
```
dmitry-timofeev: I assume 'contains an index' means 'contains some kind of Exonum table'. If that's true, I'd say 'A median of these ~~index~~values', because it's already clear where these values are stored and 'index values' sounds confusing (what do these values _index_?).

src/advanced/time.md
```diff
@@ -0,0 +1,287 @@
...
+
+- The time value is stored in the Exonum storage; hence,
+  verifying its authenticity requires additional
+  cryptographic checks from lightweight clients.
+- Exonum blocks are burdened with time oracle transactions
+  (usually one for each validator).
+
+### Choosing Approach
+
+Given the pros and cons above, **exonum-time** uses the time oracle approach
```
dmitry-timofeev: A time oracle seems to be what we want to get. Do you mean here 'time service'?

src/advanced/time.md
```diff
@@ -0,0 +1,287 @@
...
+as a more flexible one.
```
dmitry-timofeev: I might misunderstand the purpose of this document, but is it even necessary to put descriptions of various approaches and such design decisions in the user guide? As a user I want to know how to use the time oracle and what to expect from it. As an inquisitive user, I might be interested in how it works. But design decisions are beyond that interest. I believe that is redundant information in the user guide, but it might be useful as project documentation.

slowli: I think it could be worth it to reposition the "Preliminaries" section as "Possible alternatives", where the alternative (i.e., integration with consensus) would be described. This would be in line with academic research, which the article already emulates to an extent.

PR #176, opened 2018-02-15
Spellcheck documentation

slowli:
This PR introduces an automated spellchecking tool, `cspell`, which is already used in the main repository. Also, the Markdown linter is switched to a JS/Node-based one - it is better supported and easier to control together with `cspell`.

PR #178, opened 2018-02-15
Prepare for develop branch

slowli:
Exonum documentation has 2 major kinds of PRs:

- "Bugfixes" or other changes in docs that can be pushed to production (i.e., exonum.com website) right away. (Some examples are #177, #176, #175).
- Description of upcoming functionality that should be pushed to the website no earlier than it is released. (The examples could be upcoming PRs describing exonum/exonum#488, exonum/exonum#481, exonum/exonum#457, etc.)

To allow to deal with the second kind of PRs efficiently, I propose to use [a well-known `master` / `develop` branching model](http://nvie.com/posts/a-successful-git-branching-model/). Note that the problem does not exist with the core Exonum repo itself (Exonum is not pushed into production); thus, it could be an argument against moving docs into the core repository.


slowli:
@DarkEld3r would you and the core team in general be OK for such a workflow for this repo? IMO, it could simplify contributing to docs - for one, I won't hesitate to merge PRs because they may describe unreleased features.

It would be interesting to think how a similar branching model could be applied to the core repo itself, but I feel we don't have that many urgent bug fixes right now.

DarkEld3r:
@slowli I fully agree with the proposed work-flow. As for me, it isn't needed for core (right now?..) because crates.io and github (tags) allow a more simple work-flow with releases. As far as I can see, we need to handle that manually for exonum-doc.

PR #179, opened 2018-02-19
Add note about timeout parameters

nataliiakovalova:


DarkEld3r:
I propose to add a note that generally it makes sense to make `round_timeout` twice bigger than `consensus.timeout_adjuster.timeout`.

nataliiakovalova:
> I propose to add a note that generally it makes sense to make round_timeout twice bigger than consensus.timeout_adjuster.timeout.
@DarkEld3r  done

src/advanced/configuration-updater.md
```diff
@@ -122,6 +122,12 @@ config][stored_configuration] serialization. It has the following fields:
   Maximum number of transactions per block.
 - **services**: Object  
   Service-specific configuration parameters.
+  
+!!! note
+    **consensus.round_timeout** integer must be strictly larger 
+    (recommended at least 2x) than **consensus.timeout_adjuster.timeout** 
+    integer, otherwise, the consensus algorithm will stop working correctly
```
slowli: I'd suggest to split the "otherwise..." part into a new sentence to improve readability.

nataliiakovalova: @slowli Thanks. Done.

PR #181, opened 2018-03-06
Add /v1/consensus_enabled docs [ECR-719]

ozkriff:


src/advanced/node-management.md
```diff
@@ -415,6 +415,52 @@ A JSON object with the following fields:
...
+Switches consensus interaction of the node on or off.
+
+#### Parameters
+
+- **enabled**: `bool`
+
+#### Example
+
+```None
+curl --data "enabled=false" http://127.0.0.1:7780/api/system/v1/consensus_enabled
```
slowli: Huh, I thought we're geared towards using `Content-Type: application/json` for POST requests, and this example uses urlencoded forms. Could you check whether it works with `bodyparser`?

src/advanced/node-management.md
```diff
@@ -415,6 +415,52 @@ A JSON object with the following fields:
...
+
+Returns a boolean value representing if the node participates in consensus.
+
+#### Parameters
+
+None.
+
+#### Response
+
+A JSON boolean
```
slowli: Nitpick: missing period at the end of the sentence.

src/advanced/node-management.md
ozkriff: Ouch, fixed.

src/advanced/node-management.md
```diff
@@ -415,6 +415,52 @@ A JSON object with the following fields:
...
+Switches consensus interaction of the node on or off.
+
+#### Parameters
+
+- **enabled**: `bool`
+
+#### Example
+
+```None
+curl --data '{"enabled":false}' http://127.0.0.1:7780/api/system/v1/consensus_enabled
```
slowli: Sorry to bug you again, but maybe add `-H "Content-Type: application/json"` to `curl` arguments just to be absolutely sure that the call does what we expect it to do? Afaik, without specifying the header it will be set to `application/x-www-form-urlencoded`.

ozkriff: Done! :)

PR #187, opened 2018-03-27
Update transaction execution info and description of relevant HTTP APIs

slowli:
This PR describes new transaction execution mechanics introduced in v0.6.

## See also
https://wiki.bf.local/display/EXN/Transaction+status

### DIFF
--- a/misc/lint.sh
+++ b/misc/lint.sh
@@ -41,6 +41,7 @@ lint_html () {
   html5validator --root "$SITE_DIR" --show-warnings --ignore-re \
     'Illegal character in query: "\|" is not allowed' \
     '"(autocorrect|autocapitalize)" not allowed on element "input"' \
+    '"autocomplete" is only allowed when the input type is' \
     '"align" attribute on the "(td|th)" element is obsolete' \
     'Document uses.* Unicode Private Use Area';
 }
--- a/src/advanced/node-management.md
+++ b/src/advanced/node-management.md
@@ -9,7 +9,7 @@ node [local configuration](../architecture/configuration.md#local-parameters).
 
 Endpoints are divided into two types: private and public endpoints. Each
 endpoint type is hosted at a separate address, which is specified in the
-[`node.api` section](../architecture/configuration.md#nodeapi) of the local
+[`api` section](../architecture/configuration.md#api) of the local
 configuration.
 
 ## Types
@@ -125,7 +125,7 @@ address formatted as 4 octets separated by points.
 ### SerializedTransaction
 
 `SerializedTransaction` is a JSON object corresponding to
-[transaction serialization format](../architecture/transactions.md#serialization)
+[transaction serialization format](../architecture/serialization.md#message-serialization).
 
 ### TransactionLocation
 
--- a/src/advanced/node-management.md
+++ b/src/advanced/node-management.md
@@ -24,18 +24,18 @@ an optional type.
 
 ### Bool
 
-`bool` type denotes a simple boolean `true/false` value.
+`bool` type denotes a boolean value: `true` or `false`.
 
 ### Hash, PublicKey, Signature
 
 `Hash`, `PublicKey`, `Signature` types are hexadecimal strings of the
-appropriate length. `Hash` and `PublicKey` consist of 32 bytes.
-`Signature` consists of 64 bytes.
+appropriate length. `Hash` and `PublicKey` are 32 bytes (that is, 64 hex digits).
+`Signature` is 64 bytes (that is, 128 hex digits).
 
 ### PeerAddress
 
 `PeerAddress` is a string containing address in `IP:port` format. `IP` is IPv4
-address formatted as 4 octets separated by points.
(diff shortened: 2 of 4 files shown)


misc/lint.sh
```diff
@@ -41,6 +41,7 @@ lint_html () {
   html5validator --root "$SITE_DIR" --show-warnings --ignore-re \
     'Illegal character in query: "\|" is not allowed' \
     '"(autocorrect|autocapitalize)" not allowed on element "input"' \
+    '"autocomplete" is only allowed when the input type is' \
```
alexander-irbis: Здесь что-то осталось недосказано?

dmitry-timofeev:
@slowli , @DarkEld3r I have a question though — has this change in Exonum affected the format of the return values of the corresponding APIs (e.g., `/transactions/<tx_hash>`)? If so, does the API description _or_ examples need to be updated?

DarkEld3r:
@dmitry-timofeev As far as I can see, yes.

dmitry-timofeev:
@DarkEld3r I’m not sure I understand :smile:

misc/lint.sh
alexander-irbis: Is there anything left unsaid?

alexander-irbis: Is there anything left unsaid?

slowli: There are a lot of `input` types in the relevant error message (see [this Travis log](https://travis-ci.org/exonum/exonum-doc/builds/358907689)), so I've skipped them.

slowli:
@dmitry-timofeev @DarkEld3r I'll update the endpoint description within this PR. I'm thinking about removing examples altogether, or reducing them to something less brittle, a la

```javascript
{
  "type": "in-pool",
  "content": {
    "protocol_version": ...,
    "service_id": ..., // Identifier of the service processing transaction
    "message_id": ...,
    "body": ..., // JSON-encoded body
    "signature": ...
  }
}
```

dmitry-timofeev:
@slowli , I'd keep at least one or two, because a proper JSON might be easier to grasp than a spec. If you also want to reduce the amount of space they take, I'd consider making them hidden by default (if that's possible with the tool we use to generate HTML pages).

misc/lint.sh
```diff
@@ -41,6 +41,7 @@ lint_html () {
   html5validator --root "$SITE_DIR" --show-warnings --ignore-re \
     'Illegal character in query: "\|" is not allowed' \
     '"(autocorrect|autocapitalize)" not allowed on element "input"' \
```
Maria-Nosyk: > '"(autocorrect|autocapitalize)" not allowed on element "input"' 
is not allowed

slowli: This is a third-party error message, so I cannot change it :disappointed_relieved:

src/advanced/node-management.md
```diff
@@ -462,6 +334,157 @@ All explorer API endpoints share the same base path, denoted
...
+    - `"success"` denotes a successfully completed transaction
+    - `"error"` denotes a transaction that has returned an error (for example,
+      because of transaction parameters not satisfying context-dependent checks)
+    - `"panic"` denotes a transaction that has raised a runtime exception
+      (for example, attempted to divide by zero)
+
+- **status.code**: integer  
+  Error code supplied by the service developer. Only present for erroneous
+  transactions. Has service-specific meaning
+- **status.description**: string=
```
dmitry-timofeev: Is the equality sign intentional?

src/advanced/node-management.md
```diff
@@ -462,6 +334,157 @@ All explorer API endpoints share the same base path, denoted
...
+  Always equals to `committed`
+- **status**: Object  
+  [Transaction execution](../architecture/transactions.md#execute) status
+- **status.type**: `"success"` | `"error"` | `"panic"`  
+  Execution status kind:
+
+    - `"success"` denotes a successfully completed transaction
+    - `"error"` denotes a transaction that has returned an error (for example,
+      because of transaction parameters not satisfying context-dependent checks)
+    - `"panic"` denotes a transaction that has raised a runtime exception
```
dmitry-timofeev: @DarkEld3r ,

Just wondering if it is OK to use "panic" (a Rust-specific term) in the context of polyglot Exonum? The other question is whether unrecoverable errors shall be treated as recoverable.

DarkEld3r: @dmitry-timofeev
Good question. I think that we can use "panic", but not as a Rust-specific term. We can explain it as an Exonum transactions' specific term.

dmitry-timofeev: OK, thanks, and what about always-attempting-to-recover Exonum? 

On top of that, I asked some more questions on the wiki page above, we can discuss this question there too.

src/advanced/node-management.md
slowli: Yes, it means that the field is optional, as described [at the beginning of the page](https://github.com/slowli/exonum-doc/blob/5a67307ddfa5c088f9b6e4b3c0bff5f3a153abac/src/advanced/node-management.md#types).

PR #188, opened 2018-03-29
Replace transactions endpoint documentation [ECR-1002]

SerhiiOryshych:


slowli:
Eh, the corresponding changes are already covered by #187...

PR #197, opened 2018-05-04
Update request to add new peer

ivan-ochc:

### DIFF
--- a/src/advanced/node-management.md
+++ b/src/advanced/node-management.md
@@ -241,7 +241,7 @@ will not be established.
 #### Example
 
 ```None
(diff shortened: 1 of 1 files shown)


src/advanced/node-management.md
```diff
@@ -241,7 +241,7 @@ will not be established.
 #### Example
 
 ```None
-curl --data "ip=127.0.0.1:8800" http://127.0.0.1:7780/api/system/v1/peers
+curl --data '{"ip":"127.0.0.1:8800"}' http://127.0.0.1:7780/api/system/v1/peers
```
slowli: Would you mind formatting JSON in a more fancy way: `'{ "ip": "127.0.0.1:8800" }'`? It could help make the example more readable.

PR #198, opened 2018-05-04
Add missing consensus parameter

MartaDubno:
Added missing consensus configuration parameter which was present at docs.rs but missing at Exonum.com

### DIFF
--- a/src/architecture/configuration.md
+++ b/src/architecture/configuration.md
@@ -78,8 +78,11 @@ two parts:
   Timeout interval (ms) for sending a `Status` message
 - **txs_block_limit**  
   Maximum number of transactions per block
-- **max_message_len**
-  Maximum message length (in bytes)
+- **max_message_len**  
+  Maximum message length (in bytes). This parameter determines the maximum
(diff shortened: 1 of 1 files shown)


src/architecture/configuration.md
```diff
@@ -78,6 +78,8 @@ two parts:
   Timeout interval (ms) for sending a `Status` message
 - **txs_block_limit**  
   Maximum number of transactions per block
+- **max_message_len**
```
slowli: There needs to be two spaces at the end of the line for correct formatting.

src/architecture/configuration.md
```diff
@@ -78,6 +78,8 @@ two parts:
...
+  Maximum message length (in bytes)
```
slowli: Would you mind adding some details? If I understand correctly, something like this:

> The parameter decides the maximum allowed size of both consensus messages and transactions. The default value of the parameter is 1 MiB, i.e., `1024 * 1024` bytes.

Please consult developers for the sane range of the values. I *think* it's somewhere around 1 MiB - 16 MiB, but I'm not completely sure.

PR #201, opened 2018-05-07
Update "blocks in range" response example

ivan-ochc:

### DIFF
--- a/src/advanced/node-management.md
+++ b/src/advanced/node-management.md
@@ -698,11 +698,11 @@ smallest and largest heights traversed to collect at most `count` blocks.
 The JSON object of the explored block range `range` and the array `blocks` of
 the `BlockHeader` objects. The range specifies the largest and the smallest
 heights of blocks that have been traversed to collect at most `count` blocks.
-The largest height `end` equals to `latest` if provided or to the height of
+The largest height `end` equals to `latest + 1` if provided or to the height of
 the latest block in the blockchain, the smallest height `start` takes values
 in `0..latest - count + 1`. Blocks in the array are sorted in descending order
 according to their heights. Height of any block in the array is greater or
-equal than `start` and less or equal than `end`.
+equal than `start` and less than `end`.
 
 ??? example "Response Example"
     Assume the following request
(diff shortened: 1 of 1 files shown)


src/advanced/node-management.md
```diff
@@ -698,11 +698,11 @@ smallest and largest heights traversed to collect at most `count` blocks.
 The JSON object of the explored block range `range` and the array `blocks` of
 the `BlockHeader` objects. The range specifies the largest and the smallest
 heights of blocks that have been traversed to collect at most `count` blocks.
-The largest height `to` equals to `latest` if provided or to the height of
-the latest block in the blockchain, the smallest height `from` takes values
+The largest height `end` equals to `latest` if provided or to the height of
```
slowli: Isn't `end` exclusive, as implied [by using `Range`](https://github.com/exonum/exonum/blob/ca48622516f821d7f5ffe5d907953a28d5e6e5e3/exonum/src/api/public/blockchain_explorer.rs#L32) in the implementation? If yes, it should be "The largest height `end` equals to `latest + 1`", right?

src/advanced/node-management.md
```diff
@@ -698,11 +698,11 @@ smallest and largest heights traversed to collect at most `count` blocks.
...
+the latest block in the blockchain, the smallest height `start` takes values
 in `0..latest - count + 1`. Blocks in the array are sorted in descending order
 according to their heights. Height of any block in the array is greater or
-equal than `from` and less or equal than `to`.
+equal than `start` and less or equal than `end`.
```
slowli: Should probably be "less than `end`"; see the comment above.

src/advanced/node-management.md
```diff
@@ -765,7 +761,11 @@ equal than `from` and less or equal than `to`.
           "tx_count": 1000,
           "tx_hash": "94f251c0350c95024f46d26cbe0f9d2ea309e2817da4bab575fc4c571140291f"
         }
-      ]
+      ],
+      "range": {
+        "end": 100,
```
slowli: Should probably be `"end": 101`.

PR #202, opened 2018-05-10
Update installation guide for Windows

ivan-ochc:

### DIFF
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -55,10 +55,11 @@ be present in the default package lists. In this case you may need to install
 these packages from third-party PPAs, or build them from sources.
 
 ### Windows
+Install the latest version of the following packages:
+* [Visual C++ Build Tools][build_tools]
+* [PowerShell][powershell]
 
-Exonum is a little behind supporting Windows OS compared to Linux and MacOS.
-In many Windows setups, Exonum will work just fine, but there are known problems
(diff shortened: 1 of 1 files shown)


src/get-started/install.md
```diff
@@ -56,9 +56,12 @@ these packages from third-party PPAs, or build them from sources.
...
 
-Exonum is a little behind supporting Windows OS compared to Linux and MacOS.
-In many Windows setups, Exonum will work just fine, but there are known problems
-with its compilation in certain configurations. If you have encountered a problem
+Install the latest version of the following packages:
+
+- [Visual C++ Build Tools][build_tools]
+- [PowerShell][powershell]
+
+If you have encountered a problem
```
dmitry-timofeev: Isn't this redundant if we _do_ support Windows properly? We probably have references to Gitter/Github issues elsewhere, don't we?

DarkEld3r: I don't have a strong opinion here, but I tend to agree. On the other hand, we definitely don't support (test) Windows platform equally to OS X/Linux, so this can make sense.

slowli: I incline to agree with @dmitry-timofeev; this paragraph seems redundant.

ivan-ochc: @slowli, do you mean paragraph "If you have encountered a problem..."?

slowli: Yeah.

PR #204, opened 2018-05-17
Update time API [ECR-1451]

SerhiiOryshych:


src/advanced/time.md
```diff
@@ -141,13 +141,10 @@ None.
 
 #### Response
 
-Example of JSON response:
+Example of response:
```
slowli: Would you mind adding that the response is combined date and time in UTC as per [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)?

src/advanced/time.md
```diff
@@ -172,10 +169,7 @@ Example of JSON response:
 [
   {
     "public_key": "83955565ee605f68fe334132b5ae33fe4ae9be2d85fbe0bd9d56734ad4ffdebd",
-    "time": {
-      "nanos_since_epoch": 626107000,
-      "secs_since_epoch": 1516011501
-    }
+    "time": "2018-05-17T10:45:56.057753Z"
```
slowli: Same for this endpoint.

PR #226, opened 2018-06-28
Spanish welcome page

boguslavsky:


misc/cspell.json
```diff
@@ -110,7 +110,128 @@
     "validators",
     "Vertx",
     "whitelisted",
-    "whitelisting"
+    "whitelisting",
+    "código",
```
slowli: I think excluding the Spanish-language file or including a separate dictionary for it (there is probably one) would both be preferable.

boguslavsky: You are right, thanks. Spanish page excluded.

PR #229, opened 2018-07-09
Redesign the Node Management section of docs

MartaDubno:
Changes to the structure of the page according to EAN-188. The code looks a bit strange. If you have any ideas how to make the code a bit more neat, they are more than welcome. The text of the page should be more or less the same. I'm attaching screenshots of what the outcome should be.

screen shot 2018-07-09 at 6 30 22 pm 2
screen shot 2018-07-09 at 6 30 33 pm 2
screen shot 2018-07-09 at 6 30 44 pm 2


slowli:
I don't quite like the new design, but that's beside the point. More importantly, a correct way to approach EAN-188 is to use [the OpenAPI spec](https://swagger.io/specification/) to document REST API of the core / services.

MartaDubno:
This PR is no longer relevant.

PR #232, opened 2018-07-18
Add archetypeVersion option in get-started/java-binding [ECR-1882]

MakarovS:
Downside of this change - we're going to need to update this field each release for now.

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -13,6 +13,7 @@ $ mvn archetype:generate \
     -DinteractiveMode=false \
     -DarchetypeGroupId=com.exonum.binding \
     -DarchetypeArtifactId=exonum-java-binding-service-archetype \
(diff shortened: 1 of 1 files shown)


dmitry-timofeev:
> Downside of this change - we're going to need to update this field each release for now.

dmitry-timofeev:
> Downside of this change - we're going to need to update this field each release for now.

I believe the user have to specify the version explicitly if they already have this artefact locally (that was the case when we used it).

MakarovS:
> I believe the user have to specify the version explicitly if they already have this artefact locally (that was the case when we used it).

When I used it without explicitly specifying the version, even if I had some newer version of archetype in my local maven cache (0.1.1 or 0.1.2), remote and old one (0.1 from maven central) was prioritised and used instead of those.

MakarovS:
@slowli could you please take a look?

PR #233, opened 2018-07-21
Change EJB doc to v0.2

MakarovS:


MakarovS:
>  I’d suggest replacing the version in Javadoc links with “latest” (currently they point to 0.1)

You mean that we change this line to `-DarchetypeVersion=[<latest_release>](https://github.com/exonum/exonum-java-binding/releases/latest) \` ? The problem is we don't have a release yet (only pre-releases), so this link leads to just https://github.com/exonum/exonum-java-binding/releases . Though we can change 0.1 and 0.2 to release versions?

dmitry-timofeev:
No, I meant updating the version in the Javadoc links like https://exonum.com/doc/api/java-binding-core/0.1/com/exonum/binding/storage/database/MemoryDb.html from "0.1" to "latest", which is an alias for the latest uploaded version (0.2 at this moment). This way we won't have to update all the links each time we release a new version. 

I'd keep the version in the command generating a project so that it can be copied easily (i.e., `-DarchetypeVersion=0.2`).

dmitry-timofeev:
I checked, the Javadocs for 0.2 are uploaded on our server and are available as "0.2" and "latest".

PR #234, opened 2018-07-22
Use latest EJB version in Javadoc

MakarovS:
@dmitry-timofeev Do we want to do something about such links as https://github.com/exonum/exonum-java-binding/blob/v0.1/exonum-java-binding-cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/ApiControllerTest.java ? Do we want to change them to latest version manually each release? Or change `v0.1` to `master`? Or just leave it be?

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -380,25 +380,25 @@ service:
 - [Rust instruction](create-service.md)
 - [Java Binding App tutorial][app-tutorial]
 
-[abstractservice]: https://exonum.com/doc/api/java-binding-core/0.1/com/exonum/binding/service/AbstractService.html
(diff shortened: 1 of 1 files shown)


dmitry-timofeev:
> Do we want to change them to latest version manually each release? Or change v0.1 to master? Or just leave it be?

I'd go with updating them each release, because the chance we notice when the referenced files are renamed or deleted would be higher.

dmitry-timofeev:
Updated in the first PR.

slowli:
Just to clarify: do I understand correctly, that the idea is to merge #233, then this PR?

dmitry-timofeev:
> Just to clarify: do I understand correctly, that the idea is to merge #233, then this PR?

I believe they may get merged in any order for there must not be any conflicts.

dmitry-timofeev:
Fixed the conflict, @slowli , would you merge please?

PR #235, opened 2018-07-27
OpenAPI specs for API

ilammy:
This is a preview version of OpenAPI (aka Swagger) specification of node API extracted from current documentation and source code. This machine-readable format is going to replace the current informal specification of the API in documentation.

It covers the following services:

* Exonum node management
* Exonum blockchain explorer
* BTC anchoring service
* configuration update service
* time oracle service

The content is rather crude, the descriptions are not fully complete, there are some spelling/grammar issues, some examples are missing. But it should be technically accurate for Exonum 0.9.0 (as in describing all endpoints, parameters, and results).

The API is currently not rendered anywhere. You can get a preview with [online OpenAPI editor](https://editor.swagger.io), just paste raw YAML content there. Please refer to those previews and provide feedback on what could be improved from technical standpoint.

### DIFF
--- a/src/advanced/bitcoin-anchoring.yaml
+++ b/src/advanced/bitcoin-anchoring.yaml
@@ -15,13 +15,13 @@ servers:
 - url: http://example.com/api/services/btc_anchoring/v1
 
 tags:
-- name: Public endpoints
-  description: Accessible by any remote node or host.
+- name: Read Requests
+  description: Public read-only API available to everybody.
 
 paths:
   /address/actual:
     get:
-      tags: [Public endpoints]
+      tags: [Read Requests]
       summary: Get actual anchoring address
       description: >
         Returns the current anchoring BTC-address.
@@ -37,7 +37,7 @@ paths:
 
   /address/following:
     get:
-      tags: [Public endpoints]
+      tags: [Read Requests]
       summary: Get next anchoring address
       description: >
         If the change of the validators list is scheduled, returns the next anchoring address.
@@ -55,7 +55,7 @@ paths:
 
   /actual_lect:
     get:
-      tags: [Public endpoints]
+      tags: [Read Requests]
       summary: Get actual common LECT
       description: >
         Returns the [LECT](https://exonum.com/doc/advanced/bitcoin-anchoring/#lect)
@@ -72,7 +72,7 @@ paths:
 
   /actual_lect/validator:
     get:
-      tags: [Public endpoints]
+      tags: [Read Requests]
       summary: Get actual LECT for specific validator
       description: >
         Returns the actual [LECT](https://exonum.com/doc/advanced/bitcoin-anchoring/#lect)
@@ -123,7 +123,7 @@ paths:
 
   /nearest_lect:
     get:
-      tags: [Public endpoints]
+      tags: [Read Requests]
       summary: Get nearest LECT for a block
       description: >
         Returns the content of the anchoring transaction which anchors the specific block.
@@ -158,7 +158,7 @@ paths:
 
   /block_header_proof:
     get:
-      tags: [Public endpoints]
+      tags: [Read Requests]
       summary: Get cryptographic proof for a block
       description: >
         Provides cryptographic proofs for Exonum blocks
--- a/src/advanced/bitcoin-anchoring.yaml
+++ b/src/advanced/bitcoin-anchoring.yaml
@@ -234,6 +234,7 @@ components:
       description: >
         A proof of existence for an anchored or a non-anchored Exonum block at the given height.
       properties:
+        # TODO: describe BlockProof, MapProof, ListProof in schema
         latest_authorized_block:
           type: object
           description: >
--- a/src/advanced/bitcoin-anchoring.yaml
+++ b/src/advanced/bitcoin-anchoring.yaml
@@ -1,0 +1,238 @@
+openapi: 3.0.0
+info:
+  title: Anchoring Service
+  version: 0.9.0
+  description: >
+    The anchoring service is developed to increase product security
+    and provide non-repudiation for Exonum applications.
+    This service periodically publishes the Exonum blockchain block hash to the Bitcoin blockchain,
+    so that it is publicly auditable by anyone having access to the Exonum blockchain.
+    Even in the case of validators collusion, transaction history cannot be falsified.
+    The discrepancy between the actual Exonum blockchain state
+    and the one written to the Bitcoin blockchain will be found instantly.
+
+servers:
+- url: http://example.com/api/services/btc_anchoring/v1
+
+tags:
+- name: Public endpoints
+  description: Accessible by any remote node or host.
+
+paths:
+  /address/actual:
+    get:
+      tags: [Public endpoints]
+      summary: Get actual anchoring address
+      description: >
+        Returns the current anchoring BTC-address.
+      responses:
+        200:
+          description: >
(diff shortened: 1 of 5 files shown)


MartaDubno:
Looks good. I found two technical issues in the UI part of the editor, not sure if its the problem with the code or the editor. The issues are shown on the screenshot below.
- This relates to all the files and the Models sections, specifically the first 2 or 4 models in them. Once I open them I cannot close these models, the button isn't there.
- In a non-full screen mode, long descriptions and examples go beyond the grey fields where they belong.
screen shot 2018-07-31 at 5 07 07 pm 2
.

ilammy:
API spec will need an update once exonum/exonum#839 is merged.

src/advanced/blockchain-explorer.yaml
```diff
@@ -0,0 +1,528 @@
...
+                  precommits:
+                    type: array
+                    description: The list of `Precommit` messages supporting the block.
+                    items:
+                      $ref: "#/components/schemas/Precommit"
+                  txs:
+                    type: array
+                    description: The list of the transactions included into the block.
+                    items:
+                      $ref: "#/components/schemas/SerializedTransaction"
```
slowli: I think the transaction hashes are used here.

src/advanced/blockchain-explorer.yaml
```diff
@@ -0,0 +1,528 @@
...
+        - **committed:** committed transaction (in blockchain)
+
+        - **in-pool:** uncommitted transaction (in the pool of unconfirmed transactions)
+
+        - **unknown:** unknown transaction
+      enum:
+      - committed
+      - in-pool
+      - unknown
+    TransactionStatus:
```
slowli: Maybe, it could make more sense to represent the status as `oneOf` with a discriminator?

src/advanced/node-management.yaml
```diff
@@ -0,0 +1,361 @@
...
+        200:
+          description: Successfully initiated node shutdown.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/NullObject"
+
+components:
+  schemas:
+    NullObject:
```
slowli: Looks clunky, but I'm not sure if a better type description is possible (e.g., `type: "null"` doesn't seem to work).

src/advanced/time.yaml
```diff
@@ -0,0 +1,120 @@
...
+                    time: "2018-05-17T10:45:56.057753Z"
+                  - public_key: "52baa9d4c4029b925cedf1a1515c874a68e9133102d0823a6de88eb9c6694a59"
+                    time: null
+
+  /validators_times/all:
+    get:
+      tags: [Private API]
+      summary: Get timestamps of all validators
+      description: >
+        Returns the latest timestamps indicated by all validator nodes for which time is known.
```
slowli: Nit: I'd elaborate that by all validator nodes we mean both current and past ones.

src/advanced/node-management.yaml
ilammy: Yeah. For some reason OpenAPI does not have `type: null` and this seems to be the closest we can get to 'only `null` is allowed here'. Though, technically this type specification allows any object in addition to `null`.

src/advanced/blockchain-explorer.yaml
ilammy: I'm not sure. It does make sense from the rigorous point of view, but I kinda doubt that we'll get any merit from it.

`oneOf` does not seem to allow to specify which discriminator values choose which variants. It is a way to join two very different schemas into one. It is used for `KnownTransaction` here (and could have been used for `UnknownTransaction` as well), but that's because committed and uncommitted transactions have very different representations.

In this case we still have a single `TransactionStatus` type with a single purpose. The most common way of handling it should be check the `type` field for `"success"` and if it's not it then propagate information from `code` and `description`. I think we can leave it as a single type with some meaningful default values for `code` and `description` and don't bother with accurate discriminated union representation.

Though, here's how it's represented in the code:

```rust
enum TxStatus<'a> {
    Success,
    Panic { description: &'a str },
    Error { code: u8, description: &'a str },
}
```

so I guess we should translate that directly into `oneOf`. Maybe some day we'll have OpenAPI docs generated from the code. If I were the generator I would translate that `enum` as `oneOf`.

src/advanced/blockchain-explorer.yaml
ilammy: Yeah, my bad. Thanks for having hawk eyes.

This object is actually serialized from this struct:

```rust
pub struct BlockInfo {
    pub block: Block,
    pub precommits: Vec<Precommit>,
    pub txs: Vec<Hash>,
}
```

ilammy:
Added a bunch of changes to resolve issues pointed out by @slowli.

One thing that bothers me is the `maximum` values for some types. We use `u64` in Rust code so I have added appropriate limits:

```yaml
minimum: 0
maximum: 18446744073709551615  # <== 2^64 - 1
```

However, when rendered by Swagger the limit it shown as 18446744073709552000. I guess Swagger can't handle unsigned 64-bit values accurately. It's not possible to specify the limit as a string, it has to be a number.

ilammy:
@MartaDubno Uncollapsible model entries seem to be a bug in Swagger's editor.

As for the examples... Do we need to keep accurate examples in data model sections? They do not seem to word-wrap correctly. Though JSON objects in request examples wrap normally. I guess we can remove the model value examples and keep them only in requests.

ilammy:
Can we merge this please?

src/advanced/bitcoin-anchoring.yaml
```diff
@@ -0,0 +1,259 @@
...
+        then an error is returned.
+      parameters:
+        - in: query
+          name: id
+          required: true
+          description: Identifier of the validator to query for.
+          schema:
+            type: integer
+            minimum: 0
+            maximum: 4294967295
```
slowli: [Validator is represented by `u16`](https://docs.rs/exonum/0.8.0/exonum/helpers/struct.ValidatorId.html), so the upper limit should be 65535.

src/advanced/bitcoin-anchoring.yaml
```diff
@@ -0,0 +1,259 @@
...
+      responses:
+        200:
+          description: >
+            A string with hex-encoded content of the nearest Bitcoin anchoring transaction.
+          content:
+            application/json:
+              schema:
+                type: string
+                nullable: true
+                format: "^[0-9a-f]+$"
```
slowli: Are you sure this works? I was under the impression (reading the OpenAPI spec) that [ECMA-262 regexps](https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.5) should be used (i.e., ones used in JavaScript; something like `/^[0-9a-f]+$/i`). Additionally, shouldn't the property be `pattern`, rather than `format`?

src/advanced/blockchain-explorer.yaml
```diff
@@ -0,0 +1,558 @@
...
+          allOf:
+            - $ref: "#/components/schemas/Hash"
+            - description: >
+                Hash of the current
+                [Exonum state](https://exonum.com/doc/glossary/#blockchain-state)
+                after applying transactions in the block.
+        tx_count:
+          type: integer
+          minimum: 0
+          maximum: 4294967295
```
slowli: I don't think it makes sense to include `u32` or `u64` upper restrictions, tbh, especially for returned types. Additionally, `u64::max_value()` isn't even represented properly in the Swagger doc viewer.

PR #238, opened 2018-08-10
Add doc-entry for rebroadcast endpoint [ECR-2131]

VladYermakov:
Added doc entry for `v1/rebroadcast` endpoint.

### DIFF
--- a/src/advanced/node-management.md
+++ b/src/advanced/node-management.md
@@ -428,6 +428,28 @@ curl -H "Content-Type: application/json" --data 'null' http://127.0.0.1:7780/api
 ```json
 null
 ```
+### Rebroadcast
+
+```none
+POST {system_base_path}/rebroadcast
+```
+
+Rebroadcast all transactions from the pool to other validators.
+
+#### Parameters
+
+None.
+
+#### Example
+
+```none
+curl -H "Content-Type: application/json" --data 'null' http://127.0.0.1:7780/api/system/v1/rebroadcast
+```
+
+#### Response
+```json
+null
+```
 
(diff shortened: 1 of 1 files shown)


DarkEld3r:
@slowli Done: this pull request now targets `develop` branch.

src/advanced/node-management.md
```diff
@@ -429,6 +429,30 @@ curl -H "Content-Type: application/json" --data 'null' http://127.0.0.1:7780/api
 null
 ```
 
+### Rebroadcast
+
+```none
+POST {system_base_path}/rebroadcast
+```
+
+Rebroadcast all transactions from the pool to other validators.
```
dmitry-timofeev: Why would a user need this endpoint (I'd explain since this is a reference documentation)?

But really, why is that needed? Can't Exonum exchange transactions properly?

DarkEld3r: > Can't Exonum exchange transactions properly?

No, it cannot in some very specific cases. For example, we have a network with some validators which are all off-line at the moment and one auditor. A user sends a transaction to the auditor, who broadcasts it to the validators (but they are off-line). As the auditors (a validator can become a leader and propose a block with such transaction) don't have transactions exchange mechanic, this transaction will stuck forever in the pool.

> Why would a user need this endpoint (I'd explain since this is a reference documentation)?

Generally, a user doesn't. This is a temporary workaround, so I'm not sure it should be explained thoroughly.

dmitry-timofeev: I see — but not sure that I understand completely how an auditor can become a leader and propose transactions?

PR #240, opened 2018-08-13
Add a warning about broken ITs on Linux

dmitry-timofeev:
## Overview
Add a warning about broken ITs on Linux since it's unknown how much it will take to fix them.

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -327,6 +327,7 @@ changes made to `Fork` to the database state:
 
 !!! warning
     Java integration tests that use the shared native library currently work on MacOS only.
+    The Java Binding App which runs Java services, however, works fine on Linux.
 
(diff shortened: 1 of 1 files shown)


src/get-started/java-binding.md
```diff
@@ -327,6 +327,7 @@ changes made to `Fork` to the database state:
 
 !!! warning
     Java integration tests that use the shared native library currently work on MacOS only.
+    The Java Binding App which runs Java services, however, works fine on Linux.
```
MartaDubno: I'd put the conjunction at the beginning:
However, the Java Binding App...
Also, does the Java Binding App work fine on both Linux and MacOS or only on Linux? If its the first case, I suggest the following wording:
However, the Java Binding App which runs Java services, works fine on both MacOS and Linux.

PR #253, opened 2018-12-11
Updated serialization document

MartaDubno:
A page describing protobuf serialization.
Any comments on what should be added/removed/improved are welcome.


src/architecture/serialization.md
```diff
@@ -100,55 +102,83 @@ include!(concat!(env!("OUT_DIR"), "/example_mod.rs"));
...
+```rust
+let exonum_protos = get_exonum_protobuf_files_path();
+    protobuf_generate(
+        "src/proto",
+        &["src/proto", &exonum_protos],
+        "example_mod.rs",
+    );
+```
+
+After applying the `get_exonum_protobuf_files_path()` function, users can
```
dvush: I think this function is called rather than applied since there are no arguments

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+lets users automatically map
+their structures and the structures generated from `.proto` descriptions,
+providing a mechanism for validating protobuf generated data. The structures
+for `ProtobufConvert` should have the same fields as the structures in
+`.proto` files, but can contain additional validation.
+
+The
+[`exonum_derive`](https://github.com/exonum/exonum/tree/master/exonum_derive)
+crate provides the ability to use structures typical for Exonum with all
+the required validations. So when using these structure users only need to
```
Maria-Nosyk: these structures

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+[light clients](../architecture/clients.md)
+and storage of data. Protobuf is the industry accepted language-neutral and
+platform-neutral automated mechanism for serializing data.
 
 ## Usage
 
-### Communication Among Full Nodes
-
-Full nodes can both [serialize messages](#message-serialization) for sending and
+**Communication Among Full Nodes** Full nodes can both
```
slowli: Nit: I like the previous formatting with headers better, but I suppose this one makes sense too.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+`.rs` extension. The `.rs` files handle the serialization of structures
+described in them. And that is it, the described structures can then be
+serialized and deserialized in Exonum.  
+
+For example, a simple `Timestamp` structure will have following description
+in a `.proto` file:
+
+```protobuf
+message Timestamp {
+
```
slowli: Nit: unnecessary empty line.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+  // Represents seconds of UTC time since Unix epoch
+  // 1970-01-01T00:00:00Z. Must be from 0001-01-01T00:00:00Z to
```
slowli: I find this description confusing because values `0001-01-01T00:00:00Z` don't look like `int64` values.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+## Building Exonum with Protobuf Serialization
 
-**Binary presentation:** `u16` (unsigned 2-byte integer).  
-**JSON presentation:** number.
+Exonum includes the
+[`exonum_build`](https://github.com/exonum/exonum/tree/master/exonum_build)
+crate which lets users add the `protobuf_generate` function to their
+`build.rs`. This function automatically generates the `.rs` files for all the
+`.proto` files during the build process. To use `protobuf_generate`, add the
+following code to `build.rs` indicating the folder which contains the `.proto`
```
slowli: Also, `exonum_build` should be added to the project as a build dependency.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
-### Signature
+Then to access a structure from a protobuf-generated file, you need to indicate
+the name of the module with the generated files and the name of the required
+file and structure in it:
 
-[Ed25519 digital signature](https://ed25519.cr.yp.to/) over the binary
-serialization of the message (excluding the signature bytes,
-i.e., the last 64 bytes of the serialization).
+```rust
+module_name::proto_file_name::StructNameInProtoFile
```
slowli: Nit: This isn't valid Rust on its own; I'd prefer a more comprehensive example.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+files that come with the `exonum` crate - `get_exonum_protobuf_files_path()`:
 
 ```rust
-const MY_SERVICE_ID: u16 = 777;
-
-transactions! {
-    Transactions {
-        const SERVICE_ID = MY_SERVICE_ID;
+let exonum_protos = get_exonum_protobuf_files_path();
+    protobuf_generate(
```
slowli: Please de-indent.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+```
 
-        struct MessageTwoIntegers {
-            first: u64,
-            second: u64,
-        }
-    }
+After calling the `get_exonum_protobuf_files_path()` function, users can
+import protobuf descriptions from the `exonum` crate in their `.proto` files.
+For example, `import "helpers.proto;"` can be used to get access to protobuf
```
slowli: Semicolon isn't in a right place here.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
-use hex::FromHex;
-
-encoding_struct! {
-    struct Wallet {
-        pub_key: &PublicKey,
-        owner: &str,
-        balance: u64,
-    }
+#[derive(Clone, PartialEq, Eq, Ord, PartialOrd, Debug, ProtobufConvert)]
+#[exonum(pb = "proto::TransactionsRequest", crate = "crate")]
```
slowli: @dvush I thought `crate = ...` part was made optional. Also, shouldn't it be `crate = "exonum"` if we're talking about *user*-defined structs? I think it would be confusing to provide an example of a structure defined in an `exonum` crate if the way it's defined differs from a way user-defined structs would be defined.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+providing a mechanism for validating protobuf-generated data. The structures
+for `ProtobufConvert` should have the same fields as the structures in
+`.proto` files, but can contain additional validation.
+
+The
+[`exonum_derive`](https://github.com/exonum/exonum/tree/master/exonum_derive)
+crate provides the ability to use structures typical for Exonum with all
+the required validations. So when using these structures users only need to
+implement `#[derive(ProtobufConvert)]` for them. If required, users can
+implement the [`ProtobufConvert`][convert] trait for any additional structures
```
slowli: I'm afraid this sentence will be quite unclear to an average reader without additional explanation.

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
-[utf8]: https://en.wikipedia.org/wiki/UTF-8
-[rust_structs]: https://doc.rust-lang.org/book/first-edition/structs.html
-[rust_enums]: https://doc.rust-lang.org/book/first-edition/enums.html
-[cryptocurrency]: https://github.com/exonum/exonum/blob/master/examples/cryptocurrency
-[rust-slice]: https://doc.rust-lang.org/book/first-edition/primitive-types.html#slicing-syntax
-[rust]: http://rust-lang.org/
-[cargo_features]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section
-[subnormal_fp]: https://en.wikipedia.org/wiki/Denormal_number
+Note that it is required to indicate the protobuf structure to which the
+current structure refers, in the case above `proto::Connect`.
```
slowli: `proto::TransactionsRequest`?

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+```
 
-The length of the entire message serialization (including its header and signature).
+To use protobuf-generated Rust structures, users first need to create a module
+which will include the protobuf-generated files:
 
-**Binary presentation:** `u32` (unsigned 4-byte integer).  
-**JSON presentation:** (not serialized).
+```rust
+extern crate exonum;
```
slowli: Why is `extern crate exonum` included here, in the container module? I'd just leave it out. In the 2018 edition, it's not required at all, and in 2015 edition it should be placed in `lib.rs` (otherwise, we would need to write `use self::exonum::...` later).

src/architecture/serialization.md
```diff
@@ -1,533 +1,193 @@
...
+`exonum_build` also includes a function which returns the path to the `.proto`
+files that come with the `exonum` crate - `get_exonum_protobuf_files_path()`:
 
 ```rust
-const MY_SERVICE_ID: u16 = 777;
-
-transactions! {
-    Transactions {
-        const SERVICE_ID = MY_SERVICE_ID;
+let exonum_protos = get_exonum_protobuf_files_path();
```
slowli: @dvush This is off-topic, but I can't help but notice that this UX seems overly complicated.

- Are there cases in which we *lose* something from including Exonum protobuf files? If no, it could make sense to make their inclusion implicit.
- Duplicating `"src/proto"` dir looks weird.
- It could probably make sense to have a convention for the output module name if we assume `protobuf_generate` will be called once per crate.
- I think it could make sense to organize `proto` files inside the included dir as `exonum/somefile.proto`, analogous to files with well-known Protobuf types (cf. `import "google/protobuf/any.proto";`).

src/architecture/serialization.md
dvush: Yes, it is optional. This one is copy from `exonum` crate, thats why it has this parameter.

This `crate = ""crate"` attribute should be removed.

src/architecture/serialization.md
dvush: 4) I agree, and I had to [go back](https://github.com/exonum/exonum/pull/1088/commits/97a9a07be88e33fde820cfd3600e2b216e685edd) to the current version since I could not figure out how to do this on windows, `protoc_generate` uses strings to compare paths in its internals and it breaks on windows because of slashes. I know that it sounds stupid but I could not figure out easy solution to this(and we are bound by time =\)

1,2) I tried to make API similar to `protoc_generate` crate + it would make things more implicit(maybe easier to debug), but now I would agree that it makes more sense to do this implicitly.

3) I prefer current interface since it explicitly ties `include!` and `protoc_generate` (it is greppable and more transparent) and because we have examples where it is called multiple times per build rs(exonum and time-service).

src/architecture/serialization.md
dvush: Thats comment from google/well_known_types. I think to show this point comments can be removed.

src/architecture/serialization.md
dvush: Maybe something like this?
```rust
use module_name::proto_file_name::StructName;

fn deserialize(bytes: &[u8]) -> StructName {
    let mut pb = StructName::new();
    pb.merge_from_bytes(slice).unwrap();
    pb
}

fn serialize(pb: &StructName) -> Vec<u8> {
    pb.write_to_bytes().unwrap()
}
```

src/architecture/serialization.md
dvush: Ok, 4) is going to be fixed soon.

slowli: - Re 1,2: Including exonum files could be an opt-out option if we worry about that.
- Re 3: Wouldn't it be an option to create a macro (say, `include_protobuf!()`) that invokes `include!(concat!(env!("OUT_DIR"), "/mod.rs"))`? It could be used to import exonum-Protobuf glue as well 🙃

src/architecture/serialization.md
```diff
@@ -1,533 +1,201 @@
...
-
-- Owner: `String`  
-  Andrew
+```rust
+let exonum_protos = get_exonum_protobuf_files_path();
+protobuf_generate(
+    "src/proto",
+    &["src/proto", &exonum_protos],
+    "example_mod.rs",
+    );
```
slowli: Nit: unnecessary indent.

src/architecture/serialization.md
```diff
@@ -1,533 +1,201 @@
...
+
+To have validation performed only once for the whole structure, Exonum
+provides the conversion mechanism using the `ProtobufConvert` trait. This trait
+lets users automatically map
+their structures and the structures generated from `.proto` descriptions,
+providing a mechanism for validating protobuf-generated data. The structures
+for `ProtobufConvert` should have the same fields as the structures in
+`.proto` files, but can contain additional validation.
+
+The [`exonum_derive`][derive]
```
slowli: The crate is named [`exonum-derive`](https://crates.io/crates/exonum-derive).

src/architecture/serialization.md
```diff
@@ -1,533 +1,201 @@
...
+current structure refers, in the case above `proto::TransactionRequest`.
+
+[protobuf]: https://developers.google.com/protocol-buffers/docs/overview
+[proto-files]: https://github.com/exonum/exonum/tree/master/exonum/src/proto/schema/exonum
+[language]: https://developers.google.com/protocol-buffers/docs/reference/proto3-spec
+[cryptocurrency]: https://github.com/exonum/exonum/blob/master/examples/cryptocurrency/src/proto/cryptocurrency.proto
+[convert]: https://github.com/exonum/exonum/blob/master/exonum/src/proto/mod.rs
+[module]:https://github.com/exonum/exonum/tree/master/examples/cryptocurrency/src/proto
+[derive]: https://github.com/exonum/exonum/tree/master/components/derive
+[build]: https://github.com/exonum/exonum/tree/master/components/build/
```
slowli: Nit: Why not provide links to `crates.io` or `docs.rs` for `-build` and `-derive` crates?

src/architecture/serialization.md
```diff
@@ -1,533 +1,201 @@
...
+[anchoring-rs]: https://github.com/exonum/exonum-btc-anchoring/blob/master/src/proto/mod.rs#L33
```
slowli: This isn't a permanent link (especially the line number). Please refer to a specific commit or tag.

src/architecture/serialization.md
```diff
@@ -1,533 +1,201 @@
...
+their structures and the structures generated from `.proto` descriptions,
+providing a mechanism for validating protobuf-generated data. The structures
+for `ProtobufConvert` should have the same fields as the structures in
+`.proto` files, but can contain additional validation.
+
+The [`exonum_derive`][derive]
+crate provides the ability to use structures typical for Exonum with all
+the required validations. So when using these structures users only need to
+implement `#[derive(ProtobufConvert)]` for them. If required, users can
+implement the [`ProtobufConvert`][convert] trait for any other structure they
```
slowli: Nit: I'd link to `docs.rs` here.

MartaDubno: This section of docs.rs currently fails to build, so unfortunately I cannot link to it.

PR #255, opened 2018-12-13
Build all develop branches on CI

dmitry-timofeev:
Build any develop (pre-release) branches, such as 'develop-ejb-0.4' on CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ cache:
 branches:
   only:
     - master
-    # Any develop (pre-release) branches, such as 'develop-ejb-0.4'
-    - /^develop-.*$/
+    # Any develop (pre-release) branches, such as 'develop' or 'develop-ejb-0.4'
+    - /^develop(-.*)?$/
     # Tags
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,9 @@ cache:
 branches:
   only:
     - master
-    - develop
-    # tags
+    # Any develop (pre-release) branches, such as 'develop-ejb-0.4'
+    - /^develop-.*$/
+    # Tags
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 before_install:
   # Lock dependencies for branch builds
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,8 +10,9 @@ cache:
 branches:
   only:
     - master
-    - develop
-    # tags
+    # Any develop (pre-release) branches, such as 'develop-ejb-0.4'
+    - /^develop-.*$/
```
slowli: Does this change mean that `develop` branch itself is not built? Or are you proposing to get rid of it in favor of "topical" development branches? In the latter case, I'd suggest to clarify this in the branching strategy description in README.

dmitry-timofeev: That’s a bug, was not intentional :-) 

However, shall I put that `develop-x` branches are fine in PR template?

slowli: I'd say yes, but I have no complete understanding of the new branching strategy yet. Will it be clear for external contributors which of `develop-*` branches they need to target?

vitvakatu: We do not have external contributors at the moment and I doubt they appear in the nearest future. These `develop-x` branches will help us to parallel work on documentation, and everyone will know what is the branch they should target.

dmitry-timofeev: I checked out the current branching policy, and, if I understand correctly, it must be adjasted either way to work for several concurrent 'features' (where by a 'feature' I mean a set of changes for a particular release, like EJB-0.4 or Exonum 0.10). The problem currently is that you cannot merge 'not ready to be released immediately' patches to `develop`, because each feature has its own release date. If, for example, EJB-0.4 is due Jan 31, but Exonum 0.10 is due Jan 1, and some documentation changes from EJB-0.4 get into `develop`, you won't be able to publish Exonum-0.10 changes until Jan 31 because they are 'tainted' with EJB-0.4 changes. Therefore, we shall either treat each feature as a feature branch under the present model **and** do not merge them to `develop` _unless_ they can be integrated immediately — not on a particular date in the future (but doesn't that makes `develop` almost useless?); or have these long-running feature branches and merge them directly to master when the time comes.

slowli: OK, I more or less got the picture. Note that you've not brought back building `develop` branch yet 🙃

PR #263, opened 2019-01-15
Update install.md to protobuf

gmarat:
Dependencies to protobuf in v0.10.x

### DIFF
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -25,6 +25,7 @@ Exonum depends on the following third-party system libraries:
 
 - [RocksDB][rocksdb] (persistent storage)
 - [libsodium][libsodium] (cryptography engine)
+- [Protocol Buffers][protobuf] (mechanism for serializing structured data)
 
 You can find instructions how to install dependencies in various environments
 below.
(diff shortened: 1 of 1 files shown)


src/get-started/install.md
```diff
@@ -44,7 +45,7 @@ use
 
 ```shell
 apt-get install build-essential libsodium-dev libsnappy-dev \
-    librocksdb-dev pkg-config
+    librocksdb-dev pkg-config libprotobuf-dev protobuf-compiler
```
vitvakatu: Had anyone check this command? I thought that these packages are outdated in Ubuntu (and I remember some reports from other developers about compilation issues with outdated protoc)

gmarat: works fine on Ubuntu 18.04.1, checked today

gmarat: not sure about 16.04 or others, maybe they need to update ppa repositories or install protoc manually

dmitry-timofeev: > I thought that these packages are outdated in Ubuntu

On 16.04 they are, one needs to add a PPA (that we also use on Travis). Protobuf isn't the most recent there either, but it seems to work for me (but I haven't run _Exonum_ test suite with that, so can't tell for sure).

gmarat: Sorry for annoying, but will it be merged? Maybe I did something wrong or i have to do something more? Asking, because I want to understand whether to send PR in the future.

PR #264, opened 2019-01-18
Add Protobuf serialization

Maria-Nosyk:


src/get-started/light-client.md
```diff
@@ -28,165 +28,272 @@ A transaction is an atomic operation that introduces changes to the blockchain
...
+}
+
+// Timestamping transaction.
+message TxTimestamp { Timestamp content = 1; }
 ```
 
-- `content_hash` is a SHA-256 hash of some data or file to be stamped
-- `metadata` is an optional description of the data to be stamped that is
-  included into the stamp.
+When the `.proto` file is ready, generate the `*.js` file with
```
boguslavsky:
```
generate the `*.js` stub file
```

src/get-started/light-client.md
```diff
@@ -28,165 +28,272 @@ A transaction is an atomic operation that introduces changes to the blockchain
...
+
+```json
+{
+  "proto": "pbjs --keep-case -t static-module example.proto -o ./proto.js"
+}
+```
+
+#### Define a Transaction Schema
+
+Define `TransferFunds` transaction schema and its field types:
```
boguslavsky: as soon as `TransferFunds` demonstrate using of proto stubs (`.js` files), please also show how to include them on the page before defining transaction:
```
import * as proto from 'stubs.js'
```

src/get-started/light-client.md
```diff
@@ -28,165 +28,272 @@ A transaction is an atomic operation that introduces changes to the blockchain
...
 
 ```javascript
-const seed = Exonum.randomUint64()
+const TransferFunds = Exonum.newTransaction({
+   author: publicKey,
+   service_id: 128,
+   message_id: 1,
+   schema: examples.TransferFunds
```
boguslavsky: Replace to:
```
schema: proto.exonum.examples.cryptocurrency_advanced.Transfer
```

src/get-started/light-client.md
```diff
@@ -273,19 +381,32 @@ const tableRootHash = Exonum.verifyTable(data.wallet_proof.to_table,
...
-const Wallet = Exonum.newType({
-  fields: [
-    { name: 'pub_key', type: Exonum.PublicKey },
-    { name: 'name', type: Exonum.String },
-    { name: 'balance', type: Exonum.Uint64 },
-    { name: 'history_len', type: Exonum.Uint64 },
-    { name: 'history_hash', type: Exonum.Hash }
-  ]
-})
+const Wallet = Exonum.newType(examples.Wallet)
```
boguslavsky: Replace with:
```
const Wallet = Exonum.newType(proto.exonum.examples.cryptocurrency_advanced.Wallet)
```

src/get-started/light-client.md
```diff
@@ -316,34 +439,29 @@ if (typeof wallet === 'undefined') {
...
+Basically, the proof from the Timestamping Service comprises the same validation
 levels as described above. Specifically, to obtain confirmation for the
 timestamp data, the proof validates the block, the root table and the table of
 timestamps. The data on the timestamp can then be extracted from the validated
 timestamps table.
 
-Meanwhile, the proof we presently investigate contains another level that refers
-to the validation of transactions in respect of each wallet.  
+Meanwhile, the proof we presently investigate contains another level. This level
+refers to the validation of transactions in respect of each wallet (wallet
```
slowli: Is "in respect of..." a real phrase? 🙃 Maybe, something like "concerning a specific wallet" would be better?

src/get-started/light-client.md
```diff
@@ -316,34 +439,29 @@ if (typeof wallet === 'undefined') {
...
+history).  
 
 First, we obtain a proof for all transactions in the wallet. It will contain
-transaction hashes as well as statuses thereof. In our example we obtain
-information for the whole wallet transaction history, however, any suitable
-range may be selected:
+transactions hashes as well as statuses thereof. In our example we obtain
+information for the whole wallet transactions history. However, any suitable
```
slowli: Nit: I'd shorten to "we obtain the whole wallet history".

src/get-started/light-client.md
```diff
@@ -316,34 +439,29 @@ if (typeof wallet === 'undefined') {
...
-const TransactionMetaData = Exonum.newType({
-  fields: [
-    { name: 'tx_hash', type: Exonum.Hash },
-    { name: 'execution_status', type: Exonum.Bool }
-  ]
-})
 
 const transactionsMetaData = Exonum.merkleProof(
-  wallet.history_hash,
+  Exonum.uint8ArrayToHexadecimal(new Uint8Array(wallet.history_hash.data)),
```
slowli: Nit / off-topic: Isn't it possible to use `Buffer` utils for hex <-> Uint8Array conversion? IMO, this kind of reuse would be preferable to rolling out our own utils.

src/get-started/light-client.md
```diff
@@ -356,50 +474,71 @@ Otherwise, transactions cannot be verified against the proof:
...
+definition of only one transaction type. This type is a transaction of token
+issuance from our [Cryptocurrency Advanced Service][cryptocurrency-advanced].
+However, note that to perform a real check, all the transaction types of the
+service should be defined.
+
+The check of the transaction structure comprises the following steps:
 
-Next, we validate each transaction signature with `Transaction.verifySignature`
-method.
+- calculate a hash from the transaction buffer with the `Exonum.hash`
```
slowli: Nit: Shouldn't sentences in this list be capitalized?

src/get-started/light-client.md
```diff
@@ -356,50 +474,71 @@ Otherwise, transactions cannot be verified against the proof:
...
+  if (Exonum.hash(buffer) !== hash) {
+     throw new Error('Invalid transaction hash')
   }
 
-  // validate hash
-  if (Transaction.hash(transaction.body) !== transactionsMetaData[index]) {
-    throw new Error('Invalid transaction hash has been found')
+  // Serialize transaction from the debug view and compare it with the
+  // `bufferWithoutSignature` obtained above.
+  if (!Transaction.serialize(transaction.debug).every(function (el, i) {
```
slowli: Nit: wouldn't it make more sense to use an arrow function syntax here? I don't think it's supported less than `for .. of`.

src/get-started/light-client.md
Maria-Nosyk: actually, it is. and it is absolutely suitable for the case but I do not mind changing the text as you suggest

src/get-started/light-client.md
Maria-Nosyk: I do not think capitalization is necessary here. The steps we enlist are in fact a contuniation of the introductory part above them. Do we have any approved rules concerning this issue? If not, I'll remember to reflect it in our style guide in future.

src/get-started/light-client.md
boguslavsky: Thanks, looks very interesting. I will investigate if it is possible to do such converting and if so I will implement it into Light client, demos and tutorial.

src/get-started/light-client.md
boguslavsky: Make sense, done

src/get-started/light-client.md
```diff
@@ -28,165 +28,274 @@ A transaction is an atomic operation that introduces changes to the blockchain
...
 
+#### Send the Transaction
+
 Finally, send the resulting transaction into the blockchain using the built-in
 `send` method which returns a `Promise`:
 
 ```javascript
-const transactionHash = await CreateTimestamp.send(transactionEndpoint,
-  explorerBasePath, data, signature)
+const transactionHash = await CreateTimestamp.send(transactionEndpoint, data, keyPair.secretKey)
```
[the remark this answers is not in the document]
boguslavsky: 👌

src/get-started/light-client.md
```diff
@@ -28,165 +28,274 @@ A transaction is an atomic operation that introduces changes to the blockchain
...
+  Hash content_hash = 1;
+  string metadata = 2;
+}
+
+// Timestamping transaction.
+message TxTimestamp { Timestamp content = 1; }
+```
+
+When the `.proto` file is ready, generate the `*.js` stub file with
+[pbjs][pbjs-lib] library as follows:
```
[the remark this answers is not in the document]
boguslavsky: Right, description added.

src/get-started/light-client.md
slowli: The only reservation I can think of is that `Buffer` is not readily available in browser environments, but it can be easily added there by using [the `buffer` package](https://www.npmjs.com/package/buffer), which provides fully Node-compatible interface.

PR #266, opened 2019-01-21
Document standard services provided by the EJB app [ECR-2819]

MakarovS:
Document standard services provided by the EJB app.

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -240,9 +240,9 @@ at `/cryptocurrency/balance/:walletId`.
 See [documentation][vertx.io] on the possibilities of `Vert.x` used as a web
 framework.
 
-### Supported Services
+### Built-In Services
 
-Currently Java Binding supports the following services:
+Currently Java Binding includes the following built-in services:
 
 - **Configuration Update Service.**
   Although every node has its own configuration file, some settings should be
@@ -260,12 +260,18 @@ Currently Java Binding supports the following services:
   See the [*Anchoring Service*](../advanced/bitcoin-anchoring.md)
   article for more details.
 
-To enable services put `ejb_app_services.toml` file into EJB App's directory
-with the following content:
+To enable a particular service, include its name in "ejb_app_services.toml"
+configuration file into EJB App's directory with the following content:
 
 ```toml
-services = ["configuration", "btc-anchoring"]
+services = ["service-name", ...]
 ```
+
+where possible values for "service-name" are: "configuration" for Configuration
+Service and "btc-anchoring" for Anchoring Service.
+
+In case of no "ejb_app_services.toml" file, only "configuration" service will
+be activated.
 
 ### Dependencies Management
 
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -243,6 +243,7 @@ framework.
 ### Supported Services
 
 Currently Java Binding supports the following services:
+
 - **Configuration Update Service.**
   Although every node has its own configuration file, some settings should be
   changed for all nodes simultaneously. This service allows updating
@@ -255,7 +256,7 @@ Currently Java Binding supports the following services:
   to the Bitcoin blockchain with a certain time interval. The anchored data is
   authenticated by a supermajority of validators using digital signature tools
   available in Bitcoin.
-  
+
   See the [*Anchoring Service*](../advanced/bitcoin-anchoring.md)
   article for more details.
 
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -247,6 +247,7 @@ Currently Java Binding supports the following services:
   Although every node has its own configuration file, some settings should be
   changed for all nodes simultaneously. This service allows updating
   configuration through the blockchain itself.
+
   See the [*Configuration Update Service*](../advanced/configuration-updater.md)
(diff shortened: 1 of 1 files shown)


src/get-started/java-binding.md
```diff
@@ -240,6 +240,33 @@ at `/cryptocurrency/balance/:walletId`.
...
 framework.
 
+### Supported Services
+
+Currently Java Binding supports the following services:
+
+- **Configuration Update Service.**
+  Although every node has its own configuration file, some settings should be
+  changed for all nodes simultaneously. This service allows updating
+  configuration through the blockchain itself.
```
vitvakatu: True, but the main purpose of configuration service is to update blockchain configuration on-the-flight, without stopping nodes. Any other operation is possible using regular configuration files.

src/get-started/java-binding.md
```diff
@@ -240,6 +240,33 @@ at `/cryptocurrency/balance/:walletId`.
 See [documentation][vertx.io] on the possibilities of `Vert.x` used as a web
 framework.
 
+### Supported Services
+
+Currently Java Binding supports the following services:
```
bullet-tooth: what does it mean "supports"? Can we call somehow Anchoring service from our service? I would add a few more sentences about these services.

src/get-started/java-binding.md
```diff
@@ -240,6 +240,33 @@ at `/cryptocurrency/balance/:walletId`.
 See [documentation][vertx.io] on the possibilities of `Vert.x` used as a web
 framework.
 
+### Supported Services
```
bullet-tooth: this section locates inside the "Service Development" section. I'm not sure it's a good place. 
In my way, the place below "How to Run a Service" will be better.

src/get-started/java-binding.md
vitvakatu: There is no public API for Anchoring service, so we cannot `call it from our service`.

bullet-tooth: yeap. So my comment is about to add some description. Because "support" is really general here.

src/get-started/java-binding.md
```diff
@@ -240,6 +240,33 @@ at `/cryptocurrency/balance/:walletId`.
...
+  The anchoring service writes the hash of the current Exonum blockchain state
+  to the Bitcoin blockchain with a certain time interval. The anchored data is
+  authenticated by a supermajority of validators using digital signature tools
+  available in Bitcoin.
+
+  See the [*Anchoring Service*](../advanced/bitcoin-anchoring.md)
+  article for more details.
+
+To enable services put `ejb_app_services.toml` file into EJB App's directory
+with the following content:
```
dmitry-timofeev: I'd consider putting in the example a "service-name" key, and specify the keys corresponding to supported services below.

Also, I think configuration is enabled by default. If so, please specify it.

src/get-started/java-binding.md
```diff
@@ -240,6 +240,33 @@ at `/cryptocurrency/balance/:walletId`.
...
+- **Anchoring Service.**
+  The anchoring service writes the hash of the current Exonum blockchain state
+  to the Bitcoin blockchain with a certain time interval. The anchored data is
+  authenticated by a supermajority of validators using digital signature tools
+  available in Bitcoin.
+
+  See the [*Anchoring Service*](../advanced/bitcoin-anchoring.md)
+  article for more details.
+
+To enable services put `ejb_app_services.toml` file into EJB App's directory
```
dmitry-timofeev: To enable a particular service, include its name in "ejb_app_services.toml" configuration file ...

src/get-started/java-binding.md
dmitry-timofeev: I'd say "includes" some "built-in/standard" services.

src/get-started/java-binding.md
dmitry-timofeev: Also I'd clarify in the section name that those aren't just any services, but built-in services.

src/get-started/java-binding.md
```diff
@@ -240,6 +240,41 @@ at `/cryptocurrency/balance/:walletId`.
...
+```toml
+services = ["service-name"]
+```
+
+where possible values for `service-name` are:
+
+- `configuration` for Configuration Service.
+- `btc-anchoring` for Anchoring Service.
+
+In case of no `ejb_app_services.toml` file, only Configuration Service will
```
MartaDubno: In case there is no `ejb_app_services.toml` file, only the Configuration Update Service...

src/get-started/java-binding.md
Maria-Nosyk: How about rephrasing the second sentence as follows:
> The present service allows updating global configuration parameters of the network without stopping the nodes. The changes are agreed upon through the consensus mechanism.

??

src/get-started/java-binding.md
```diff
@@ -240,6 +240,42 @@ at `/cryptocurrency/balance/:walletId`.
 See [documentation][vertx.io] on the possibilities of `Vert.x` used as a web
 framework.
 
+### Built-In Services
```
dmitry-timofeev: I'd agree with Oleg that this section location is out of place. I think it must be placed either after "How to Run a Service" on the same level or _in_ "How to Run a Service" on the level below. I'd prefer the former.

PR #284, opened 2019-02-21
Adopted testkit documentation for exonum v0.10.x

aleksuss:


src/get-started/test-service.md
```diff
@@ -177,22 +186,22 @@ failures:
 
 > **Test:** `test_transfer`
 
-Let’s test a transfer between two wallets. First, we need to create the testkit
-and the wallets:
+Let’s test a transfer between two wallets. First, we need to create the
+testkit and the wallets:
```
Maria-Nosyk: Let’s test a transfer between two wallets. To do this, first, we need to initialize the testkit. Then we need to create the wallets and transfer funds between them. As per the code of the Cryptocurrency Service, the wallets are created with the initial balance set to `100`. 

The mentioned three transactions will be included into the block:

src/get-started/test-service.md
```diff
@@ -229,9 +239,9 @@ Namely, the `create_block_with_transactions` call is replaced with
 
 ```rust
```
Maria-Nosyk: Line 235:
Let’s test the case when Alice sends a transaction...

src/get-started/test-service.md
```diff
@@ -60,21 +60,30 @@ contain the tests for transaction business logic.
 ### Imports
 
 The created file is executed separately from the service code, meaning that we
```
slowli: I believe this sentence makes little sense with the new code excerpt: in Rust-2018, it's no longer necessary to import crates in the source code (which the section previously illustrated), so it looks weird. I'd suggest to remove the sentence and the following code sample entirely.

src/get-started/test-service.md
```diff
@@ -123,19 +132,19 @@ persisted by the blockchain.
 #[test]
 fn test_create_wallet() {
     let mut testkit = init_testkit();
-    let (pubkey, key) = crypto::gen_keypair();
+    let (pubkey, key) = gen_keypair();
     testkit.create_block_with_transactions(txvec![
-        TxCreateWallet::new(&pubkey, "Alice", &key),
+        TxCreateWallet::sing("Alice", &pubkey, &key),
```
slowli: `sign`?

src/get-started/test-service.md
```diff
@@ -379,5 +391,5 @@ logic and HTTP API.
 [integration-testing]: http://doc.crates.io/manifest.html#integration-tests
```
slowli: Nit: Could you please update the link to point to `https://doc.rust-lang.org/cargo/reference/manifest.html#integration-tests`?

src/get-started/test-service.md
```diff
@@ -123,19 +132,19 @@ persisted by the blockchain.
 #[test]
 fn test_create_wallet() {
     let mut testkit = init_testkit();
-    let (pubkey, key) = crypto::gen_keypair();
+    let (pubkey, key) = gen_keypair();
     testkit.create_block_with_transactions(txvec![
-        TxCreateWallet::new(&pubkey, "Alice", &key),
+        TxCreateWallet::sing("Alice", &pubkey, &key),
     ]);
     let wallet = {
```
slowli: Nit / off-topic: I wonder If additional scope is really necessary here; seems redundant.

src/get-started/test-service.md
```diff
@@ -59,22 +59,16 @@ contain the tests for transaction business logic.
...
-The created file is executed separately from the service code, meaning that we
-need to import the service crate along with **exonum** and **exonum-testkit**:
-
-```rust
-extern crate exonum;
-extern crate exonum_cryptocurrency as cryptocurrency;
-#[macro_use] extern crate exonum_testkit;
-```
-
 Just like with the service itself, we then import the types we will use:
```
slowli: IMO, this sentence could be improved after edits. I'd suggest to change to "We need to import the types we will use."

src/get-started/test-service.md
```diff
@@ -59,22 +59,16 @@ contain the tests for transaction business logic.
...
-use exonum_testkit::{TestKit, TestKitBuilder};
-// Import datatypes used in tests from the crate where the service is defined.
+use assert_matches::assert_matches;
+use exonum::api::Error;
+use exonum::messages::{to_hex_string, RawTransaction, Signed};
+use exonum_crypto::{gen_keypair, PublicKey, SecretKey};
+use exonum_testkit::{txvec, ApiKind, TestKit, TestKitApi, TestKitBuilder};
+use serde_json::json;
+// Imports datatypes used in tests from the crate where the service is defined.
 use cryptocurrency::schema::{CurrencySchema, Wallet};
```
slowli: Shouldn't the crate name be changed to `exonum_cryptocurrency`? I'd also group the `use` statements here as well.

aleksuss: I'm not sure about renaming to `exonum_cryptocurrency`. As I understand, in this tutorial we create integration tests for just created cryptocurrency example, don't we ?

aleksuss: But let it be. Not principle.

src/get-started/test-service.md
```diff
@@ -287,14 +291,14 @@ because it may be needed to sign other transactions authorized by the wallet
...
-`get_wallet` is implemented as
+`get_wallet` is implemented as:
 
 ```rust
-fn get_wallet(&self, pubkey: &PublicKey) -> Wallet {
-    self.inner.get(
-        ApiKind::Service("cryptocurrency"),
-        &format!("v1/wallet/{}", pubkey.to_string()),
-    )
+fn get_wallet(&self, pub_key: &PublicKey) -> Wallet {
```
slowli: Nit: It seems that `get_wallet` and `assert_no_wallet` [take `PublicKey` by value in the example](https://github.com/exonum/exonum/blob/ba6834334b43e6ad197b1523f093f2d98d69716b/examples/cryptocurrency/tests/api.rs#L259), which makes sense since it's a `Copy` type.

PR #285, opened 2019-02-21
A service testing documentation was adopted for exonum v0.10.x

aleksuss:


src/advanced/service-testing.md
```diff
@@ -37,21 +37,17 @@ After that, create file `tests/transactions.rs` with the content similar
...
 #[test]
 fn test_my_tx() {
     // Create simple testkit network.
     let mut testkit = TestKitBuilder::validator()
-        .with_service(MyService::new())
+        .with_service(MyService)
         .create();
     // Create transaction.
-    let tx = MyTransaction::new(...);
+    let tx = MyTransaction::sign(...);
```
slowli: Are `sign` methods auto-generated for transactions? If not (which seems to be the case), it could be confusing to use them without warning the reader that they would need to implement these methods manually, or use `Message::sign_transaction()`.

I'd suggest adding the following text after the code sample:

Here, we assume that the service developer has implemented `sign` constructor for `MyTransaction` struct, which returns a signed transaction, `Signed<RawTransaction>`. This method is not implemented automatically; it could be replaced with more verbose, but universal

```rust
use exonum::messages::Message;
let tx = Message::sign_transaction(
    MyTransaction { /* fields */ },
    MyService::ID, // service identifier
    public_key,    // ...of the signer
    &secret_key,   // ...of the signer
);
```

PR #290, opened 2019-02-25
Add note about Windows support

vitvakatu:
Windows support is considered experimental at the moment. Due to lack of proper testing on different Windows versions and in different environments.

### DIFF
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -65,6 +65,10 @@ these packages from third-party PPAs, or build them from sources.
 
 ### Windows
 
+!!! note
+    Windows support is experimental. In case of any issues, please create an
+    issue on GitHub: https://github.com/exonum/exonum/issues/new
+
 Install the latest version of the following packages:
 
 - [Visual C++ Build Tools][build_tools]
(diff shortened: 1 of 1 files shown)


src/get-started/install.md
```diff
@@ -65,6 +65,10 @@ these packages from third-party PPAs, or build them from sources.
 
 ### Windows
 
+!!! note
+    Windows support is experimental. In case of any issues, please create an
+    issue on GitHub: https://github.com/exonum/exonum/issues/new
```
slowli: I'd insert a link here instead:

```markdown
...In case of any issues, please [create an issue on GitHub][win-issue].

[win-issue]: https://github.com/exonum/exonum/issues/new?title=Windows+support+problem
```

Note that you can set the initial title of the issue via `title` query param. Maybe, there's a better title than "Windows support problem"?

dmitry-timofeev: > Maybe, there's a better title than "Windows support problem"?

Only an issue template with a zillion of fields to enter: Windows OS version, MSVC version, … :upside_down_face:

PR #302, opened 2019-03-22
Update Timestamping Tutorial to v0.10

Maria-Nosyk:


src/get-started/timestamping-tutorial.md
```diff
@@ -70,35 +80,30 @@ where we indicate the external libraries required for our service.
...
     extern crate log;
-    extern crate serde;
     #[macro_use]
     extern crate serde_derive;
-    extern crate serde_json;
 
     pub mod api;
+    pub mod proto;
     pub mod schema;
     pub mod transactions;
```
ivan-ochc: Seems, that we may remove `mod` section from this snippet since we mention about `mods` further.

src/get-started/timestamping-tutorial.md
```diff
@@ -465,36 +625,22 @@ structures:
...
 
-- the `PublicApi` structure which enables us to implement public API endpoints
-  for our service:
-
-  ```rust
-  pub struct PublicApi;
-  ```
-
-Now we are ready to define the three endpoints required for our service:
+### API for Transactions
```
ivan-ochc: Should we mention that `handle_timestamp`, `handle_timestamp_proof` and `wire` must be implemented within `PublicApi`?

aleksuss: I guess it's redundant because in the code below we can see
```rust
.public_scope()
```
method.

PR #304, opened 2019-03-22
Add info about installing librocksdb from our PPA [ECR-3040]

vitvakatu:
On some old Ubuntu versions RocksDB shows unpredictable behavior for EJB, and all our CI is done using RocksDB 5.17. To avoid bugs, we should mention this fact in the docs.

### DIFF
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -44,8 +44,9 @@ For distributives with `deb`-based package managers (such as Debian or Ubuntu),
 use
 
 ```shell
+add-apt-repository ppa:exonum/rocksdb
 apt-get install build-essential libsodium-dev libsnappy-dev \
-    pkg-config libprotobuf-dev protobuf-compiler
+    librocksdb5.17 pkg-config libprotobuf-dev protobuf-compiler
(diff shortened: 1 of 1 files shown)


src/get-started/install.md
```diff
@@ -45,7 +45,7 @@ use
 
 ```shell
 apt-get install build-essential libsodium-dev libsnappy-dev \
-    librocksdb-dev pkg-config libprotobuf-dev protobuf-compiler
+    pkg-config libprotobuf-dev protobuf-compiler
```
dmitry-timofeev: But wouldn't it be better to just replace this with  
```
add-apt-repository ppa:exonum/rocksdb
apt-get install build-essential libsodium-dev libsnappy-dev \
    librocksdb5.17 pkg-config libprotobuf-dev protobuf-compiler
```

for simplicity? Users _always_ need the RocksDB, don't they?

dmitry-timofeev: Here it is removed completely, and the wording below suggests one shall only install it on _some_ Ubuntu distributions.

PR #305, opened 2019-03-26
Update definitions to v0.10

Maria-Nosyk:

### DIFF
--- a/src/glossary.md
+++ b/src/glossary.md
@@ -83,14 +83,14 @@ parties
 
 ## Blockchain Explorer
 
-An [application][explorer] for monitoring blocks and transactions in Exonum
-blockchain. The explorer allows obtaining the following information and
-executing the following actions via its endpoints:
+A [module][explorer] that provides REST API for monitoring blocks and
+transactions in Exonum blockchain. The explorer allows obtaining the following
+information and executing the following actions via its endpoints:
 
 - information on a specified range of blocks
 - information on a block at a specific height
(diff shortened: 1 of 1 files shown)


src/glossary.md
```diff
@@ -506,17 +515,19 @@ Transactions are ordered and grouped into [blocks](#block) in the course of
 same
 order on all [full nodes](#full-node) in the blockchain network.
 
-In Exonum, transactions are a subtype of [service endpoints](#service-endpoint).
+In Exonum, transactions are a subtype of
+[blockchain explorer endpoints](#blockchain-explorer). It allows sending
+transactions into the pool of unconfirmed transactions and broadcasting them to
```
aleksuss: ... of uncommitted transactions ... ?

src/glossary.md
```diff
@@ -84,6 +81,19 @@ one of the design goals of a blockchain is achieving accountability of the
 parties
 (e.g., internal and external auditors, regulators and end users of the system).
 
+## Blockchain Explorer
+
+An [application][explorer] for monitoring blocks and transactions in Exonum
```
aleksuss: `application` is not suitable definition I guess. An application it's something holistic. But blockchain explorer is a part of exonum framework that provides some functionality like monitoring blocks and transactions. Therefore I'd name it module or even REST API for monitoring blocks and transactions. 
```A module that provides REST APT for monitoring blocks and transactions...``` 
maybe smth like this.

PR #310, opened 2019-04-11
Update generate-config and run commands

Maria-Nosyk:

### DIFF
--- a/src/get-started/timestamping-tutorial.md
+++ b/src/get-started/timestamping-tutorial.md
@@ -880,13 +880,17 @@ performed in the directory containing `Cargo.toml`.
 
    ```shell
    exonum-demo-timestamping generate-config example/common.toml \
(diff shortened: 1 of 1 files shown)


src/get-started/timestamping-tutorial.md
```diff
@@ -879,13 +880,13 @@ performed in the directory containing `Cargo.toml`.
 
    ```shell
    exonum-demo-timestamping generate-config example/common.toml \
-       example/pub_1.toml example/sec_1.toml --peer-address 127.0.0.1:6331
+       example/pub_1.toml example/sec_1.toml --peer-address 127.0.0.1:6331 -n
```
aleksuss: What is `-n` in the end of the line ? I can't remember that we have a such parameter.

ivan-ochc: We have to add path to each **consensus** and **service** files

PR #323, opened 2019-05-07
Require rocksdb (temporarily) [ECR-3169]

dmitry-timofeev:
It has been discovered that the packaging
procedure does not produce RocksDB-independent
binary (java_bindings library), hence the app users
must install RocksDB on macOS.

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -23,7 +23,7 @@ Follow these steps:
   For Mac OS, you can use [Homebrew][brew-install] as follows:
 
   ```bash
-  brew install libsodium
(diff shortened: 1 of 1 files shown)


src/get-started/java-binding.md
```diff
@@ -23,7 +23,7 @@ Follow these steps:
   For Mac OS, you can use [Homebrew][brew-install] as follows:
 
   ```bash
-  brew install libsodium
+  brew install libsodium rocksdb
```
dmitry-timofeev: L16 is specifically *not* updated till ECR-3169 is resolved.

PR #331, opened 2019-06-11
Updated info about `exonum-app` [ECR-3074]

skletsun:
- $PATH should contain path to `installed_app_path/bin' folder
- removed all info about $EXONUM_HOME
- mentioned that $JAVA_HOME can be used to configure JVM used by the app

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -35,34 +35,16 @@ Follow these steps:
   unzip /path/to/downloaded/exonum-java-0.6.0-release.zip
   ```
 
-- Create an environment variable `EXONUM_HOME` pointing at this
-  location. You should also add an entry to the `PATH` variable.
+- Add the `bin` directory from this location to the `PATH` variable:
 
   ```bash
-  export EXONUM_HOME=~/bin/exonum-java-0.6.0-release
-  export PATH="$PATH:~/bin/exonum-java-0.6.0-release"
+  export PATH="$PATH:~/bin/exonum-java-0.6.0-release/bin"
   ```
 
   Debug builds are also available on the [release page][github-releases] and
   can be installed similarly.
 
-- This step is not necessary during installation, but is required to configure
-  the JVM to use by the application. Add a path to your JVM library to the
-  `LD_LIBRARY_PATH` environment variable. You can use the following script:
-
-  <!-- cspell:disable -->
-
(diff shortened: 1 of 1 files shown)


src/get-started/java-binding.md
```diff
@@ -35,34 +35,16 @@ Follow these steps:
...
-    2>&1 > /dev/null |\
-    grep 'java.home' |\
-    awk '{print $3}')}"
-  LIBJVM_PATH="$(find ${JAVA_HOME} -type f -name libjvm.* | xargs -n1 dirname)"
-
-  export LD_LIBRARY_PATH="${LIBJVM_PATH}"
-  ```
-
-  <!-- cspell:enable -->
+- By default, the application tries to determine the location of JVM from the already installed Java, but you can configure the JVM to use with the JAVA_HOME environment variable.
```
vitvakatu:
```suggestion
- By default, the application tries to determine the location of JVM from the already
  installed Java, but you can configure the JVM to use with the `JAVA_HOME`
  environment variable.
```

dmitry-timofeev:
I think EXONUM_HOME shall be kept in the section on testing (currently it mentions EXONUM_**JAVA**:

> `$EXONUM_JAVA` environment variable should point at [add: Exonum Java] installation location, ~~as specified in How to Run a Service section~~ [no longer specified there].

src/get-started/java-binding.md
```diff
@@ -35,34 +35,18 @@ Follow these steps:
...
-    2>&1 > /dev/null |\
-    grep 'java.home' |\
-    awk '{print $3}')}"
-  LIBJVM_PATH="$(find ${JAVA_HOME} -type f -name libjvm.* | xargs -n1 dirname)"
-
-  export LD_LIBRARY_PATH="${LIBJVM_PATH}"
-  ```
-
-  <!-- cspell:enable -->
+- By default, the application tries to determine the location of JVM from the already
```
dmitry-timofeev: Updated ECR-3048 to move this to configuration section once we have it.

I think the description is not accurate: _by default_ (or, with higher precedence), Exonum uses the JVM located in `JAVA_HOME`, if it is set. If it is not set, it uses the JVM on the `PATH`.

dmitry-timofeev:
It must probably be closed as obsolete.

PR #332, opened 2019-06-19
Fix broken markdown

bullet-tooth:


vitvakatu:
Can't understand why it works in [node management](https://github.com/exonum/exonum-doc/blob/master/src/advanced/node-management.md#known-uncommitted-transaction)

PR #336, opened 2019-06-26
adjust dependencies for ver 0.11

Maria-Nosyk:

### DIFF
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -52,8 +52,8 @@ apt-get install libsodium-dev libsnappy-dev \
 for installation. Meanwhile, as these packages reduce the build time, thus,
 their installation is highly recommended.
 
-You should also indicate the routes to these packages in the environment
-variables:
+You should also set the paths to libraries of these packages in the
+environment variables:
 
 ```shell
 export ROCKSDB_LIB_DIR=/usr/lib/x86_64-linux-gnu
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -49,8 +49,8 @@ apt-get install libsodium-dev libsnappy-dev \
 ```
 
 `libsodium-dev`, `libsnappy-dev` and `librocksdb-dev` are optional libraries
-for installation. Meanwhile, as these packages reduce the build time, thus,
-their installation is highly recommended.
+for installation. Meanwhile, as these packages reduce the build time, their
+installation is highly recommended.
 
 You should also set the paths to libraries of these packages in the
 environment variables:
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -44,9 +44,7 @@ For distributives with `deb`-based package managers (such as Debian or Ubuntu),
 use
 
 ```shell
-add-apt-repository ppa:exonum/rocksdb
-apt-get install build-essential libsodium-dev libsnappy-dev \
-    librocksdb5.17 pkg-config libprotobuf-dev protobuf-compiler
+apt-get install libssl-dev pkg-config
 ```
 
 For `protobuf` installation add the following dependencies:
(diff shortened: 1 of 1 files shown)


src/get-started/install.md
```diff
@@ -44,9 +44,7 @@ For distributives with `deb`-based package managers (such as Debian or Ubuntu),
 use
 
 ```shell
-add-apt-repository ppa:exonum/rocksdb
-apt-get install build-essential libsodium-dev libsnappy-dev \
-    librocksdb5.17 pkg-config libprotobuf-dev protobuf-compiler
+apt-get install libssl-dev pkg-config
```
dmitry-timofeev: Don't we need libsodium or librocksdb or snappy :confused: ? 

CC @vitvakatu

aleksuss: We do need

ivan-ochc: I don't have any problem with Exonum compilation on ubuntu 16.04 (docker run --rm -it ubuntu:16.04) with only such packages are installed.

dmitry-timofeev: Probably, except the build time? @aleksuss are they considered as _recommended_ packages (= to speed up the build using pre-built system libraries) or _required_ — but @ivan-ochc has determined it is likely not the case? @Maria-Nosyk Shall we communicate this distinction in the docs or just say — install them all?

aleksuss: @dmitry-timofeev @ivan-ochc Actually these packages could build under the hood. But they could be recommended cause could decrease build time of Exonum. I think it would be good to mention this in the manual. Any thoughts ?

Maria-Nosyk: As far as I can understand, it is desirable to install all the packages. Otherwise the current reduced list of dependencies will hinder the installation process. We can leave the initial list of packages and put an explanation afterwards. It will say that some packages are optional but are required for swift installation process.

aleksuss: @Maria-Nosyk exactly 👍🏻

src/get-started/install.md
```diff
@@ -44,9 +44,22 @@ For distributives with `deb`-based package managers (such as Debian or Ubuntu),
...
+apt-get install libsodium-dev libsnappy-dev \
+    librocksdb-dev libssl-dev pkg-config
 ```
 
+`libsodium-dev`, `libsnappy-dev` and `librocksdb-dev` are optional libraries
+for installation. Meanwhile, as these packages reduce the build time, thus,
+their installation is highly recommended.
+
+You should also indicate the routes to these packages in the environment
+variables:
```
aleksuss: You should also set the paths to libraries of these packages in the environment variables:

src/get-started/install.md
```diff
@@ -44,9 +44,22 @@ For distributives with `deb`-based package managers (such as Debian or Ubuntu),
...
 
+`libsodium-dev`, `libsnappy-dev` and `librocksdb-dev` are optional libraries
+for installation. Meanwhile, as these packages reduce the build time, their
+installation is highly recommended.
+
+You should also set the paths to libraries of these packages in the
+environment variables:
+
+```shell
+export ROCKSDB_LIB_DIR=/usr/lib/x86_64-linux-gnu
```
dmitry-timofeev: Rocksdb seems to be in /usr/lib:
```
$ find /usr/lib/ -iname *rocks*
/usr/lib/librocksdb.so.5
/usr/lib/librocksdb.so.5.17.2
/usr/lib/librocksdb.so.5.17
/usr/lib/librocksdb.so
```

https://github.com/exonum/exonum-java-binding/blob/35fb3b5c2fc4372ec110602f52ef088e612ba805/.travis.yml#L34-L35

dmitry-timofeev: That's on Ubuntu Linux, Mac might have different paths

aleksuss: It depends on package config as well.

dmitry-timofeev: BTW, we no longer offer/maintain our PPA, do we?

aleksuss: For fresh distributives it's unnecessary, I suppose.

PR #345, opened 2019-07-16
Remove rocksdb from requirements as it will be linked statically. [ECR-3324]

vitvakatu:
Starting from release 0.7.0 we will link RocksDB statically on Mac.

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -31,8 +31,7 @@ deployment.
   ```
 
 - Install [Libsodium][libsodium] as the necessary runtime dependency.
-  For Mac OS also install RocksDB library:
-  
+
 ??? example "Linux (Ubuntu)"
     ```bash
     sudo apt-get update && sudo apt-get install libsodium-dev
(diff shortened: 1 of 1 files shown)


src/get-started/java-binding.md
```diff
@@ -31,16 +31,15 @@ deployment.
   ```
 
 - Install [Libsodium][libsodium] as the necessary runtime dependency.
-  For Mac OS also install RocksDB library:
-  
+
 ??? example "Linux (Ubuntu)"
     ```bash
     sudo apt-get update && sudo apt-get install libsodium-dev
```
dmitry-timofeev: @dip56 @vitvakatu if the Rocksdb is linked dynamically on Linux and is not distributed with the app, why isn't it required to run a packaged application?

vitvakatu: Well, that's because our packaged app is linked statically on Linux 🤷‍♂ That's because on Linux it's not necessary to set ROCKSDB_STATIC=1, it's enough to set `ROCKSDB_LIB_DIR`

dmitry-timofeev: If statically, then the instructions are good.

PR #350, opened 2019-07-26
Add info on Noise protocol

Maria-Nosyk:

### DIFF
--- a/src/advanced/network.md
+++ b/src/advanced/network.md
@@ -58,6 +58,24 @@ full node has been really authorized by a supermajority of validators.
 Full nodes use the
 [Exonum binary serialization format](../glossary.md#binary-serialization)
 over TCP to communicate with each other.
+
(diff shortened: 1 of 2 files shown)


src/advanced/network.md
```diff
@@ -58,6 +58,24 @@ full node has been really authorized by a supermajority of validators.
...
+Noise Protocol protects Exonum against a number of potential vulnerabilities,
+for example, traffic sniffing between nodes.
+
+!!! warning
+    Nodes compiled with previous versions of Exonum will not connect to nodes
+    updated to 0.8 and further.
+
+!!! note
+    The current implementation of connection encryption is not complete: the
+    Noise channels are encrypted, but not authenticated.
```
pavel-mukhanov: Since version 0.9 connections are authenticated. We need to specify a remote public key when establishing a connection and receiver node will authenticate sender via that key.

PR #351, opened 2019-07-29
Update EJB installation guide for 0.7

vitvakatu:
Removes requirement of setting `PATH` and `LD_LIBRARY_PATH` variables. Valid only for 0.7 with [this fix](https://github.com/exonum/exonum-java-binding/pull/1040)

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -69,7 +69,7 @@ follow the instructions in [Contribution Guide][how-to-build].
   location.
 
   ```bash
-  export EXONUM_HOME=~/bin/exonum-java-0.7.0-release
+  export EXONUM_HOME=~/bin/exonum-java-0.7.0-release # or take the path from Brew output
   export PATH="$PATH:$EXONUM_HOME/bin"
   ```
 
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -69,7 +69,8 @@ follow the instructions in [Contribution Guide][how-to-build].
   location.
 
   ```bash
-  export EXONUM_HOME=~/bin/exonum-java-0.7.0-release # or take the path from Brew output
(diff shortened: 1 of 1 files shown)


src/get-started/java-binding.md
```diff
@@ -66,34 +66,16 @@ follow the instructions in [Contribution Guide][how-to-build].
 ### After Install
 
 - Create an environment variable `EXONUM_HOME` pointing at installation
-  location. You should also add an entry to the `PATH` variable.
+  location.
 
   ```bash
   export EXONUM_HOME=~/bin/exonum-java-0.7.0-release
-  export PATH="$PATH:$EXONUM_HOME"
```
dmitry-timofeev: Shan't we move it to the manual installation section because it is not required for brew (or is it?)?

vitvakatu: You need to set EXONUM_HOME using Brew, but the app will be in PATH automatically.

vitvakatu: So maybe it's worth moving to manual section, but then I need to move setting `EXONUM_HOME` to Brew section as well

dmitry-timofeev: Let's keep it as is. The only thing is that the example shows the path from the section on  manual installation, but brew will install it somewhere else. We might clarify that with a comment:

```
export EXONUM_HOME=... # Or take the path from brew output
```

dmitry-timofeev:
Does it supersede #331 ?

vitvakatu:
Actually it would be good to add info about setting JAVA_HOME

vitvakatu:
And fix annoying travis

vitvakatu:
> Actually it would be good to add info about setting JAVA_HOME

Nevermind, can't find appropriate section for it. Let's add it in ECR-3048.

PR #359, opened 2019-09-09
Bump EJB version to 0.8.0 and update Javadoc links

MakarovS:
Bump EJB version to 0.8.0 and update Javadoc links.

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -27,7 +27,7 @@ deployment.
   ```bash
   mkdir -p ~/bin
   cd ~/bin
-  unzip /path/to/downloaded/exonum-java-0.7.0-release.zip
+  unzip /path/to/downloaded/exonum-java-0.8.0-release.zip
   ```
 
 - Install [Libsodium][libsodium] as the necessary runtime dependency.
@@ -75,7 +75,7 @@ follow the instructions in [Contribution Guide][how-to-build].
 
   ```bash
   # The path is provided in after-install message in case of Homebrew
-  export EXONUM_HOME=~/bin/exonum-java-0.7.0-release
+  export EXONUM_HOME=~/bin/exonum-java-0.8.0-release
   export PATH="$PATH:$EXONUM_HOME/bin"
   ```
 
@@ -92,7 +92,7 @@ mvn archetype:generate \
     -DinteractiveMode=false \
     -DarchetypeGroupId=com.exonum.binding \
     -DarchetypeArtifactId=exonum-java-binding-service-archetype \
-    -DarchetypeVersion=0.7.0 \
+    -DarchetypeVersion=0.8.0 \
     -DgroupId=com.example.myservice \
     -DartifactId=my-service \
     -Dversion=1.0.0
@@ -104,7 +104,7 @@ You can also use the interactive mode:
 mvn archetype:generate \
     -DarchetypeGroupId=com.exonum.binding \
     -DarchetypeArtifactId=exonum-java-binding-service-archetype \
-    -DarchetypeVersion=0.7.0
+    -DarchetypeVersion=0.8.0
 ```
 
 The build definition files for other build systems (e.g., [Gradle](https://gradle.org/))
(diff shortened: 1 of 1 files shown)


vitvakatu:
On line 64 we say, that the user needs to follow the step in `After Install` section after installing Homebrew package. In recent version you no more need to update your `PATH` in case of Homebrew installation, so I wonder if we should remove this requirement. Though the user still needs `EXONUM_JAVA`.

Ideas on how to rephrase it?

src/get-started/java-binding.md
```diff
@@ -764,7 +764,7 @@ HTTP client.
 An example of API service tests can be found in
 [`ApiControllerTest`][apicontrollertest].
 
-[apicontrollertest]: https://github.com/exonum/exonum-java-binding/blob/ejb/v0.7.0/exonum-java-binding/cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/ApiControllerTest.java
+[apicontrollertest]: https://github.com/exonum/exonum-java-binding/blob/ejb/v0.8.0/exonum-java-binding/cryptocurrency-demo/src/test/java/com/exonum/binding/cryptocurrency/ApiControllerTest.java
 [auditor]: ../glossary.md#auditor
 [in-pool]: https://exonum.com/doc/version/0.11/advanced/consensus/specification/#pool-of-unconfirmed-transactions
```
dip56: I think we should use `0.12` version

src/get-started/java-binding.md
```diff
@@ -1008,39 +1008,39 @@ For using the library just include the dependency in your `pom.xml`:
...
-[transactionconvererter]: https://exonum.com/doc/api/java-binding/0.7.0/com/exonum/binding/core/service/TransactionConverter.html
+[transactionconvererter]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/service/TransactionConverter.html
 [vertx-web-docs]: https://vertx.io/docs/vertx-web/java/#_basic_vert_x_web_concepts
 [maven-install]: https://maven.apache.org/install.html
-[cryptofunctions-ed25519]: https://exonum.com/doc/api/java-binding/0.7.0/com/exonum/binding/common/crypto/CryptoFunctions.html#ed25519--
-[createpublicapi]: https://exonum.com/doc/api/java-binding/0.7.0/com/exonum/binding/core/service/Service.html#createPublicApiHandlers-com.exonum.binding.service.Node-io.vertx.ext.web.Router-
+[cryptofunctions-ed25519]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/common/crypto/CryptoFunctions.html#ed25519--
+[createpublicapi]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/service/Service.html#createPublicApiHandlers-com.exonum.binding.service.Node-io.vertx.ext.web.Router-
 [transaction-result]: https://docs.rs/exonum/0.11/exonum/blockchain/struct.TransactionResult.html
```
dip56: the same `0.12`

src/get-started/java-binding.md
MakarovS: Yes, changed this to `../advanced/consensus/specification/#pool-of-unconfirmed-transactions` yesterday, but can't push my changed because of branch protection settings at the moment :(

MakarovS:
> On line 64 we say, that the user needs to follow the step in After Install section after installing Homebrew package. In recent version you no more need to update your PATH in case of Homebrew installation, so I wonder if we should remove this requirement. Though the user still needs EXONUM_JAVA.
Ideas on how to rephrase it?

Add
```
Note that setting PATH variable is not needed in case of Homebrew installation.
```
?
@vitvakatu what do you think?

vitvakatu:
Looks good to me

src/get-started/java-binding.md
```diff
@@ -174,11 +175,11 @@ A set of named collections constitute a *service scheme*. For convenient access
 to service collections you can implement a factory of service collections.
 
 *The state of the service in the blockchain* is determined by the list of
-root hashes of its Merkelized collections. Said root hashes, when aggregated
-with root hashes of other Merkelized collections in the blockchain, form a
+index hashes of its Merkelized collections. Said index hashes, when aggregated
```
dmitry-timofeev: is determined by the list of hashes. [any hashes — nothing requires to use index hashes] Usually, it is comprised of the hashes of its Merklezied collections. State hashes of each service are aggregated in a single blockchain state hash, ...

PR #371, opened 2019-11-14
Add a notice about an order of transactions

aleksuss:


src/advanced/node-management.md
```diff
@@ -659,7 +659,7 @@ A JSON object with the following fields:
 - **precommits**: Array<Hash\>  
   List of hashes of the precommit messages supporting the block.
 - **txs**: Array<TransactionInfo\>  
-  List of the transactions included into the block.
+  List of the transactions included into the block in order as its executed.
```
Maria-Nosyk:
```suggestion
  List of the transactions included into the block. The transactions are arranged according to the order of their execution.
```

PR #378, opened 2019-12-04
Update the documentation for dynamic services:

dmitry-timofeev:
- Changes in service implementation
- Updated transaction processing documentation
- Added Schema example
- Added Configuration and Reconfiguration

The documentation on Testkit, Node configuration,
service deploy and instantiation are updated
separately.

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -1143,6 +1143,7 @@ For using the library just include the dependency in your `pom.xml`:
 [blockchain]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/blockchain/Blockchain.html
 [brew-install]: https://docs.brew.sh/Installation
 [build-description]: https://github.com/exonum/exonum-java-binding/blob/ejb/v0.8.0/exonum-java-binding/service-archetype/src/main/resources/archetype-resources/pom.xml
+[configurable]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/service/Configurable.html
 [env_logger-docs]: https://docs.rs/env_logger/0.6.2/env_logger/#enabling-logging
 [env_logger-home]: https://crates.io/crates/env_logger
 [Exonum-services]: ../architecture/services.md
@@ -1162,6 +1163,7 @@ For using the library just include the dependency in your `pom.xml`:
 [node-submit-transaction]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/service/Node.html#submitTransaction(com.exonum.binding.transaction.RawTransaction)
 [slf4j-home]: https://www.slf4j.org/
 [standardserializers]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/common/serialization/StandardSerializers.html
+[standard-supervisor-rustdoc]: https://docs.rs/exonum-supervisor/0.13.0-rc.2/exonum_supervisor/
 [storage-indices]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/storage/indices/package-summary.html
 [time-oracle]: ../advanced/time.md
 [transaction]: https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/transaction/Transaction.html
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -41,7 +41,7 @@ deployment.
     ```bash
     sudo apt-get update && sudo apt-get install libsodium-dev
     ```
-  
+
 ??? example "Mac OS"
     ```bash
     brew install libsodium
@@ -173,8 +173,8 @@ See [Serialization](#serialization) for details.
 
 A set of named collections constitute a *service schema*. Collections shall be
 defined in a unique to all service instances namespace to avoid name collisions;
-that is usually achieved by adding a prefix to the name of each collection, 
-declared in the service (e.g., an assigned service name: `"timestamping.timestamps"`). 
+that is usually achieved by adding a prefix to the name of each collection,
+declared in the service (e.g., an assigned service name: `"timestamping.timestamps"`).
 For convenient access to service collections you can implement a factory
 of service collections.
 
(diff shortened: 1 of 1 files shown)


src/get-started/java-binding.md
```diff
@@ -162,17 +162,21 @@ See [Serialization](#serialization) for details.
...
     }
     ```
 
-A set of named collections constitute a *service scheme*. For convenient access
-to service collections you can implement a factory of service collections.
+A set of named collections constitute a *service schema*. Collections shall be
+defined in a unique to all service instances namespace to avoid name collisions;
+that is usually achieved by adding a prefix to the name of each collection, 
+declared in the service (e.g., an assigned service name or numeric id:
+"timestamping.timestamps"). For convenient access to service collections
```
slowli: Nit: `"timestamping.timestamps"` should use code formatting. Also, since `Prefixed` accesses is already merged in the code, I'd suggest to mention service name only instead of service name or numeric ID.

src/get-started/java-binding.md
```diff
@@ -225,9 +277,9 @@ rules – see the corresponding section of our [documentation][transactions].
 Transactions are transmitted by external service clients to the framework as
 [Exonum messages][transactions-messages].
 A transaction message contains a header with the identifying information,
-such as an ID of the service this transaction belongs to and a transaction ID
-within that service; a payload containing transaction parameters;
-a public key of the author and a signature that authenticates them.
+such as a numeric ID of the service this transaction belongs to 
+and a transaction ID within that service; a payload containing transaction
```
slowli: > transaction ID

This identifier is called method ID in the core. Do you think "transaction ID" is better because transactions are not (yet) methods in EJB?

src/get-started/java-binding.md
```diff
@@ -225,9 +277,9 @@ rules – see the corresponding section of our [documentation][transactions].
 Transactions are transmitted by external service clients to the framework as
 [Exonum messages][transactions-messages].
 A transaction message contains a header with the identifying information,
-such as an ID of the service this transaction belongs to and a transaction ID
-within that service; a payload containing transaction parameters;
-a public key of the author and a signature that authenticates them.
+such as a numeric ID of the service this transaction belongs to 
+and a transaction ID within that service; a payload containing transaction 
+parameters; a public key of the author and a signature that authenticates them.
```
slowli: Nit: List formatting for transaction contents would be more readable.

src/get-started/java-binding.md
```diff
@@ -303,15 +351,18 @@ The `Transaction#execute` method describes the operations that are applied to th
 current storage state when the transaction is executed. Exonum passes
 an [execution context][transaction-execution-context] as an argument,
 which provides a `Fork` – a view that allows performing modifying
```
slowli: Nit: List formatting here would be more readable.

src/get-started/java-binding.md
```diff
@@ -303,15 +351,18 @@ The `Transaction#execute` method describes the operations that are applied to th
...
+about the corresponding transaction message: its SHA-256 hash that uniquely 
+identifies it, and the author’s public key.
 A service schema object can be used to access data collections of this service.
 
 Also, `Transaction#execute` method may throw `TransactionExecutionException`
 which contains a transaction error report. This feature allows users to notify
 Exonum about an error in a transaction execution whenever one occurs.
 It may check the preconditions before executing a transaction and either
 accepts it or throws an exception that is further transformed into an Exonum
+<!-- fixme: Shall we reference the new core enum (ExecutionStatus + ExecutionError)?
```
slowli: Protobuf message *could* make sense, but I'm not sure what's a good way to link to it (other than to the source code).

src/get-started/java-binding.md
```diff
@@ -326,6 +377,27 @@ for the given transaction. An observable result is the one that affects the
...
+
+Exonum delivers this notification to implementations
+of [`Service#beforeCommit`][service-before-commit] method. The method will see
+the state of the database after all transactions have been applied.
+It may make any changes to the service state, which will be included
+in the block. As transactions, the implementations must produce
+the same observable result on all nodes of the system. 
+If exceptions occur in this handler, Exonum will roll back any changes made
+to the persistent storage in it. 
+<!-- todo: Shall we specify that: -->
```
slowli: This is going to change soon: exonum/exonum#1576. The question where the result is saved is quite logical, so it could make sense to specify that it's discarded for now.

src/get-started/java-binding.md
```diff
@@ -326,6 +377,27 @@ for the given transaction. An observable result is the one that affects the
 blockchain state hash: a modification of a collection that affects the service
 state hash, or an execution exception.
 
+<!-- todo: Shall we put under "Transactions" (as a modifying operation)
+or under "Blockchain Events" (as a handler of 'before commit' event)? -->
+
+#### Block Commit Handler
```
slowli: This name is confusing given that the "Blockchain Events" section talks about *another* block commit handler. I'd suggest to use a term not including "commit" (since there is no guarantee that the block will be committed when the handler is run). Of course, it's complicated by the fact that the hook itself is called `before_commit` (not for long though, see exonum/exonum#1577). The new name - "after transactions" [handler] - looks more understandable to me.

src/get-started/java-binding.md
```diff
@@ -395,6 +466,62 @@ at `/api/services/cryptocurrency/balance/:walletId`.
 See [documentation][vertx-web-docs] on the possibilities of `Vert.x` used as a web
 framework.
 
+### Configuration
+
+#### Initialization and Initial Configuration
+
+If a service needs to initialize its persistent state when it is started,
+potentially, using some configuration parameters, it shall
```
slowli: Nit: I'm not sure I understand this usage of "shall"; looks posh. I'd say something like "Service can initialize its persistent state on its start using `Service#initialize`".

src/get-started/java-binding.md
```diff
@@ -395,6 +466,62 @@ at `/api/services/cryptocurrency/balance/:walletId`.
...
+is similar to the one for consensus configuration. 
+This protocol includes the following steps:
+1. a proposal of a new configuration; 
+2. verification of its correctness; 
+3. approval of the proposal; and
+4. application of the new approved configuration.
+
+The protocol of the _proposal_ and _approval_ steps is determined by the installed 
+supervisor service.
+The service must implement _verification_ and _application_ of the parameters by
```
slowli: > The service **must** implement ...

Services don't need to implement `Configurable` interface if they don't want to; you explicitly mention this below. Hence, "must" doesn't look appropriate.

src/get-started/java-binding.md
```diff
@@ -181,6 +185,54 @@ in each committed block. When using `AbstractService`, the hash list must be def
 in the schema class that implements [`Schema`][schema] interface; when implementing
 `Service` directly – in the service itself.
 
+!!! note "Example of a Service Schema with a single Merklized collection"
```
slowli: We use "Merkelized" spelling (don't ask why 🙃).

src/get-started/java-binding.md
```diff
@@ -395,6 +466,62 @@ at `/api/services/cryptocurrency/balance/:walletId`.
...
+This protocol includes the following steps:
+1. a proposal of a new configuration; 
+2. verification of its correctness; 
+3. approval of the proposal; and
+4. application of the new approved configuration.
+
+The protocol of the _proposal_ and _approval_ steps is determined by the installed 
+supervisor service.
+The service must implement _verification_ and _application_ of the parameters by
+implementing [`Configurable`][configurable] interface.
```
MakarovS: Add links for [configurable] and [standard-supervisor-rustdoc] (or corresponding TODOs)?

src/get-started/java-binding.md
dmitry-timofeev: We may use the newer term in this documentation, but in all the interfaces in upcoming 0.9 it is still named transaction id (incl. their Javadocs).

src/get-started/java-binding.md
dmitry-timofeev: We could actually link it, because our 'common' module builds the message classes.

But the bigger problem is that most protobuf _messages_ have bad documentation (and hence the generated classes), see https://jira.bf.local/browse/ECR-3731

dmitry-timofeev: That's how it looks like:
зображення

src/get-started/java-binding.md
```diff
@@ -326,6 +377,27 @@ for the given transaction. An observable result is the one that affects the
...
+or under "Blockchain Events" (as a handler of 'before commit' event)? -->
+
+#### Block Commit Handler
+
+Services can receive a notification after all transactions in a block has been
+processed but before the block has been committed, allowing inspecting
+the changes made by the transactions and modifying any service state.
+
+Exonum delivers this notification to implementations
+of [`Service#beforeCommit`][service-before-commit] method. The method will see
```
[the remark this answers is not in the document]
dmitry-timofeev: I mean "will see [if accesses] the state of the database ...". Might be changed to 'observe', but 'see' seems simpler.

src/get-started/java-binding.md
```diff
@@ -395,6 +466,62 @@ at `/api/services/cryptocurrency/balance/:walletId`.
...
+<!-- TODO: Link to the sections on starting new services -->
+
+Exonum invokes this method when a new service instance is added
+to the blockchain. Administrators, initiating this action,
+may pass some initialization parameters. A service is responsible for 
+validation of these parameters. If they are incorrect, it shall throw
+an Exception, which will abort the start of this service instance.
+ 
+A service shall also update its _persistent_ state based on these parameters.
+It shall not derive the `Service` _object_ state from them, as
```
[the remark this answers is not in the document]
dmitry-timofeev: I am not sure how to best put it. "no state in Service object" is a more general statement (it is broader than initial configuration) and might not be a strict requirement, but is simpler to understand. "Do not derive Service object state from initial parameters" is a more focused statement and _is_ a requirement, but is more complex.

@slowli , which one shall we use? Or is there a clearer communication?

PR #382, opened 2019-12-12
Use correct rocksdb package in installation instructions

vitvakatu:
Exonum now uses RocksDB 6.2, not 5.17. Using correct package is important when building EJB App from sources.

### DIFF
--- a/src/get-started/install.md
+++ b/src/get-started/install.md
@@ -45,8 +45,9 @@ use
 
 ```shell
 add-apt-repository ppa:exonum/rocksdb
+apt-get update
 apt-get install build-essential libsodium-dev libsnappy-dev libssl-dev \
-    librocksdb5.17 pkg-config clang-7 lldb-7 lld-7
+    librocksdb6.2 pkg-config clang-7 lldb-7 lld-7
 ```
 
 For `protobuf` installation add the following dependencies:


src/get-started/install.md
```diff
@@ -45,8 +45,9 @@ use
 
 ```shell
 add-apt-repository ppa:exonum/rocksdb
+apt-get update
 apt-get install build-essential libsodium-dev libsnappy-dev libssl-dev \
-    librocksdb5.17 pkg-config clang-7 lldb-7 lld-7
+    librocksdb6.2 pkg-config clang-7 lldb-7 lld-7
```
dmitry-timofeev: Is that correct that exonum uses a different one:

https://github.com/exonum/exonum/blob/0793790a8a37102f4a709dd109792f3118d51662/.travis.yml#L29

aleksuss: Here we assume that we install rocksdb from our source: `ppa:exonum/rocksdb`.

vitvakatu: Exonum does not depend on used RocksDB version much, as EJB do. But perhaps it would be wise to use the same library in Core Travis as well.

aleksuss: https://github.com/exonum/exonum/pull/1613

PR #388, opened 2020-01-25
Update proofs [ECR-4160] [ECR-4162]

slowli:
This PR updates "Merkelized list" article and adds a "Merkelized map" article.

See: https://jira.bf.local/browse/ECR-4160, https://jira.bf.local/browse/ECR-4162


src/advanced/merkelized-map.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+where
+
+- `HashTag::MapBranchNode = 4` is the domain separation tag for MPT nodes.
+- `key_path` is the 256-bit MPT path obtained by transforming the entry key.
+- `value_hash` is the domain-separated hash of the entry value
+  with tag `HashTag::Blob = 0`.
+
+`0x01` and `0x00` are single bytes present for historic reasons.
```
slowli: Don't quite like this.

src/advanced/merkelized-map.md
```diff
@@ -0,0 +1,233 @@
...
+      bytes key = 1;
+      oneof maybe_value {
+        bytes value = 2;
+        google.protobuf.Empty no_value = 3;
+      }
+    }
+
+    // Proof path and corresponding hash value.
+    message MapProofEntry {
+      // `proof_path` 34-byte serialization; see MerkleDB docs
```
slowli: Don't quite like this either.

src/advanced/merkelized-map.md
popzxc: Totally agree. I even created a task for that ([ECR-3485 Merkledb: Unify keys hashing in MapProofIndex](https://jira.bf.local/browse/ECR-3485)) a while ago, but had no time to implement it.

src/advanced/merkelized-map.md
popzxc: If we will always use a compressed form for hashing, I think it would worth storing entries in compressed form as well.

src/advanced/merkelized-list.md
```diff
@@ -98,153 +104,188 @@ Key | Height | Index | Value
...
-5. Custom functions to compute `val` hash for each individual entity type are
-  required on client. Each function should construct a byte array from
-  `ValueJson` fields using [Exonum serialization spec](../architecture/serialization.md)
-  and compute the hash of `val` according to [2](#hashing-rules).
+One could think of proofs as of Merkle trees pruning. That is, a proof is
+produced by “collapsing” some intermediate nodes in the Merkle tree.
+Another point of view – from the light client perspective – is that a proof
+is essentially a limited view of a list, for which the Merkle tree is
+constructed. This view allows to calculate the hash of the whole list and
+contains some of its elements.
```
aleksuss: I'd rephrase this sentence. 
```
This view allows to calculate the hash of the whole list and recognise that some of its elements are really in the list.
```
???

PR #389, opened 2020-01-26
Describe service interaction [ECR-4158]

slowli:
This PR adds a *Service Interaction* article which describes interaction among Exonum services.

See: https://jira.bf.local/browse/ECR-4158

### DIFF
--- a/src/advanced/service-interaction.md
+++ b/src/advanced/service-interaction.md
@@ -173,8 +173,8 @@ Top-level calls within a block are authorized as follows:
 - [Hooks](../architecture/services.md#hooks) executing before and
   after transactions are authorized by a special *blockchain* authority
 
-Information about call auth has a forward-compatible uniform representation
-(cf. addresses in Ethereum).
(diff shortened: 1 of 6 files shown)


src/advanced/service-interaction.md
```diff
@@ -0,0 +1,218 @@
...
+!!! warning
+    Interaction among services is considered an experimental feature
+    as of Exonum 1.0. Its interfaces and supported features may change
+    in the future.
+    Additionally, support of interaction may be limited in some environments,
+    e.g., for [Java services](../glossary.md#java-service).
+
+## Problem Overview
+
+The problem solved by the techniques described in this article
```
popzxc: IMHO this introduction sentence is an example of non-necessary officialese.

I'd rephrase it to be simpler, like:


> Given multiple services in the blockchain, some of them may want to interact with other services.
The interaction can include either reading the data of other service, or modifying its data using a service public interface.

PR #392, opened 2020-01-29
Update config [ECR-4151]

slowli:
This PR updates the *Configuration* article.

See: https://jira.bf.local/browse/ECR-4151


src/architecture/configuration.md
```diff
@@ -6,118 +6,109 @@
...
+    separated into an independent executable).
+
+Global configuration may be changed using [the supervisor service](../advanced/supervisor.md),
+and local configuration [by editing the configuration file](#changing-configuration).
+When the blockchain is being created, the initial global configuration
+is supplied from a file, too, since the blockchain is absent at this point.
+
+## Genesis Configuration
+
+**Genesis configuration** is the global configuration necessary to start
```
aleksuss:
```suggestion
**Genesis configuration** is the global configuration that necessary to start
```

src/architecture/configuration.md
```diff
@@ -6,118 +6,109 @@
...
 
 This categorization holds for both core and service parameters.
+The following table shows three possible parameter categories.
+
+| Scope   | Core | Anchoring service |
+|---------|------|-------------------|
+| Global  | Validators’ public keys | Anchoring address |
+| Local   | Validator’s private key | (N/A) |
+
+Note that there is no unified mechanism for local service configuration.
```
aleksuss:
```suggestion
Note that there is no unified mechanism for a changing local service configuration.
```
?

src/architecture/configuration.md
```diff
@@ -6,118 +6,109 @@
...
+and local configuration [by editing the configuration file](#changing-configuration).
+When the blockchain is being created, the initial global configuration
+is supplied from a file, too, since the blockchain is absent at this point.
+
+## Genesis Configuration
+
+**Genesis configuration** is the global configuration necessary to start
+a new node. While nodes update the global configuration using
+the supervisor service, a node need to get the genesis configuration from
+a trusted source. Indeed, at the point the genesis config is used, the node
```
aleksuss:
```suggestion
a trusted source. Indeed, at this point the genesis config is used, the node
```

src/architecture/configuration.md
popzxc: I think it's okay without "that" here. Or, if not, it should be "that is".

src/architecture/configuration.md
popzxc: This is not correct, IMHO. We are talking about **the** point *when* config will be used.

PR #397, opened 2020-02-04
Update test cryptocurrency service doc [ECR-4180]

aleksuss:

### DIFF
--- a/src/get-started/test-service.md
+++ b/src/get-started/test-service.md
@@ -99,13 +99,13 @@ To perform testing, we first need to create a network emulation – the eponymou
 Since `TestKit` will be used by all tests, it is natural to move its
 constructor to a separate function:
 
-<!-- markdownlint-disable MD013 -->
 ```rust
 fn init_testkit() -> TestKit {
-    TestKit::for_rust_service(CryptocurrencyService, INSTANCE_NAME, INSTANCE_ID, ())
+    TestKit::for_rust_service(
+        CryptocurrencyService, INSTANCE_NAME, INSTANCE_ID, ()
+    )
 }
 ```
-<!-- markdownlint-enable -->
 
 That is, we create a network emulation, in which there is a single validator
 node, and a single `CurrencyService`. `TestKit` supports
@@ -306,7 +306,7 @@ fn get_wallet(&self, pub_key: PublicKey) -> Wallet {
 ```
 
 That is, the method performs an HTTP GET request with the URL address
-corresponding to a service with the specified name and a `v1/wallet/…`
+corresponding to a service with the specified name and a `v1/wallet`
 path within the service API. When we created the service,
 we [defined](create-service.md#wire-api) that invoking such a request
 would return information about a specific wallet.
@@ -335,8 +335,9 @@ fn assert_no_wallet(&self, pub_key: PublicKey) {
 ```
 
 Note that this method uses the `unwrap_err` method instead of
-`unwrap`. While `unwrap` will panic if the returned value will contain
-an error, what is expectedly in this case.
+`unwrap`. While `unwrap` will panic if the returned value is
+erroneous, `unwrap_err` acts in the opposite way, panicking if
+the response does not contain an error.
 
 ### Creating Blocks
 
--- a/src/get-started/test-service.md
+++ b/src/get-started/test-service.md
@@ -58,7 +58,7 @@ We cover both kinds of testing in separate sections below.
 
 ## Testing Transaction Logic
 
-Let’s create [`tests/tx_logic.rs`][tests-tx_logic.rs], a file which will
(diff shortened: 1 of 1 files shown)


src/get-started/test-service.md
```diff
@@ -58,27 +58,26 @@ We cover both kinds of testing in separate sections below.
 
 ## Testing Transaction Logic
 
-Let’s create [`tests/tx_logic.rs`][tests-tx_logic.rs], a file which will
+Let’s create [`src/tx_tests.rs`][tests-tx_logic.rs], a file which will
```
slowli: Since transaction logic tests are no longer *integration* tests, this should be reflected in the "Preparing for Integration Testing" section; the section itself should probably be renamed to "Preparing for Testing." Also, versions of crates at the end of the section should be updated.

src/get-started/test-service.md
```diff
@@ -100,13 +99,13 @@ To perform testing, we first need to create a network emulation – the eponymou
 Since `TestKit` will be used by all tests, it is natural to move its
 constructor to a separate function:
 
+<!-- markdownlint-disable MD013 -->
 ```rust
 fn init_testkit() -> TestKit {
-    TestKitBuilder::validator()
-        .with_service(CurrencyService)
-        .create()
+    TestKit::for_rust_service(CryptocurrencyService, INSTANCE_NAME, INSTANCE_ID, ())
```
slowli: Could make sense to define constants beforehand and explain what they mean; currently, they appear out of nowhere here.

src/get-started/test-service.md
```diff
@@ -248,7 +246,8 @@ executed.
 We should check that Alice did not send her tokens to nowhere:
 
 ```rust
-assert_eq!(alice_wallet.balance, 100);
+assert!(try_get_wallet(&testkit, &alice.public_key()).is_none());
```
slowli: `try_get_wallet` and `get_wallet` functions need to be defined beforehand; currently, they appear out of nowhere.

src/get-started/test-service.md
```diff
@@ -289,27 +288,26 @@ impl CryptocurrencyApi {
...
         .get("v1/wallet")
         .unwrap()
 }
 ```
 
 That is, the method performs an HTTP GET request with the URL address
-corresponding to a service
-[with the specified name](../architecture/services.md#service-identifiers)
-and a `v1/wallet/…` path within the service API. When we created the service,
+corresponding to a service with the specified name and a `v1/wallet/…`
```
slowli: This was an error previously, but `…` in the path is bogus; the path should be simply `v1/wallet`.

src/get-started/test-service.md
```diff
@@ -322,19 +320,23 @@ with a specified public key:
...
     let err = self
         .inner
-        .public(ApiKind::Service("cryptocurrency"))
+        .public(ApiKind::Service(INSTANCE_NAME))
         .query(&WalletQuery { pub_key })
         .get::<Wallet>("v1/wallet")
         .unwrap_err();
 
-    assert_matches!(err, Error::NotFound(ref body) if body == "Wallet not found");
+    assert_eq!(err.http_code, api::HttpStatusCode::NOT_FOUND);
```
slowli: It could be worth explaining here how errors returned by Rust services are structured.

src/get-started/test-service.md
```diff
@@ -322,19 +320,23 @@ with a specified public key:
...
 }
 ```
 
-Note that this method uses the `TestKitApi::get_err` method instead of
-`TestKitApi::get`. While `get` will panic if the returned response is erroneous
-(that is, has non-20x HTTP status), `get_err` acts in the opposite way,
-panicking if the response *does not* have a 40x status.
+Note that this method uses the `unwrap_err` method instead of
+`unwrap`. While `unwrap` will panic if the returned value will contain
+an error, what is expectedly in this case.
```
slowli: This sentence doesn't read right, IMO. I'd suggest to replace it with something like

> While `unwrap` will panic if the returned value is erroneous, `unwrap_err` acts in the opposite way, panicking if the response *does not* contain an error.

src/get-started/test-service.md
```diff
@@ -100,13 +99,13 @@ To perform testing, we first need to create a network emulation – the eponymou
 Since `TestKit` will be used by all tests, it is natural to move its
 constructor to a separate function:
 
+<!-- markdownlint-disable MD013 -->
```
slowli: I'd suggest to format `TestKit::for_rust_service` in several lines instead; presently, it creates a horizontal scrollbar. Also, the `<!-- markdownlint-enable -->` comment is the reason of the failing CI build (`mkdocs` inserts a bogus `</p>` tag after it for some reason).

PR #402, opened 2020-02-12
Describe new configuration parameters for standard configuration message [ECR-4212]

vitvakatu:


src/get-started/java-binding.md
```diff
@@ -1375,6 +1375,35 @@ YAML dictionary. The Instance Configuration plugin also requires a list of
 additional parameters. In the present example the parameters refer to the
```
dmitry-timofeev: Would it make sense to rewrite the sentence "The Instance Configuration plugin also requires a list of
additional parameters" to be more concrete, as these are not additional, but one of the primary parameters:

"The Instance Configuration plugin supports several configuration formats: 
* some standard text formats: JSON, properties, plain text;
* protocol buffers."

and then define two sub-sections:

- Text Configuration Formats

- Protocol Buffers Configuration Format (already added)

vitvakatu: I did it, please take a look

vitvakatu:
> Shan't we target some ejb develop branch instead of master?

Oh lol, I created new branch, but forgot to set it as target

src/get-started/java-binding.md
```diff
@@ -1371,9 +1371,47 @@ service sources and place it inside some known directory.
...
+
+- Standard text formats: text, JSON, [java.util.Properties].
+- Arbitrary Protocol Buffers messages.
+
+#### Text Configuration Formats
+
+Requires the following parameters:
+
+- `format`. Describes the format of the configuration string. Possible values
+  are: `text`, `json` and `properties` (see [java.util.Properties]).
```
bullet-tooth: shouldn't they be upper-cased i.e. TEXT?

src/get-started/java-binding.md
```diff
@@ -1371,9 +1371,47 @@ service sources and place it inside some known directory.
...
+- `from_file`. A path to a file containing configuration string. May be absolute
+  or relative to the working directory.
+
+```yaml
+instances:
+  timestamping:
+    artifact: timestamping
+    config:
+      format: "properties"
+      value: "time_service_name=time"
```
bullet-tooth: it'll be great to see in the example how to specify a few pairs 🙃

dmitry-timofeev: > how

Certainly not inline :-)

src/get-started/java-binding.md
vitvakatu: It would be inconsistent with the overall syntax of the configuration file. However, it's a good idea to allow it in plugin.

src/get-started/java-binding.md
vitvakatu: By using `.properties` file, yes

bullet-tooth: great then, our doc looks like a usual tutorial :-)

https://www.youtube.com/watch?v=MAlSjtxy5ak

vitvakatu: Well, I can replace `properties` by `text` in this example. Would it work?

dmitry-timofeev: But do we need to document how to specify a string path to a file, or to write _service-specific_ configuration (in JSON, properties, plain text)?

If `value` is useful only for a single arguments (I have no idea how to specify multi-line string in TOML), we may use value_from_file as an example instead.

PR #405, opened 2020-02-14
Update Python version & fix versioning script

slowli:
This PR updates Python to 3.7 (2.x is no longer supported) and fix the versioning script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "description": "JS dependencies for Exonum docs",
   "dependencies": {},
   "devDependencies": {
+    "broken-link-checker": "^0.7.8",
     "cspell": "^2.1.1",
     "fs-extra": "^8.1.0",
     "git-interface": "^2.0.4",
@@ -17,6 +18,7 @@
     "cspell": "cspell --config=misc/cspell.json src/**/*.md",
     "md": "markdownlint --config=misc/markdownlint.json src/**/*.md",
     "versioned": "node versioned.js",
+    "links": "blc -r -e http://localhost:8000/",
     "test": "./misc/lint.sh all"
   },
   "repository": {
--- a/versioned.js
+++ b/versioned.js
@@ -46,7 +46,8 @@ const generateVersionedDocs = async (versions) => {
     const configFile = `./version/${version}.yml`
     const newSrc = `./version/src_${version}`
     versionedMkdocs.extra.versions = extraVersions
-    versionedMkdocs.docs_dir = newSrc
+    versionedMkdocs.docs_dir = `./src_${version}`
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "description": "JS dependencies for Exonum docs",
   "dependencies": {},
   "devDependencies": {
+    "broken-link-checker": "^0.7.8",
```
dmitry-timofeev: _broken_ link checker 🙃 ?

versioned.js
```diff
@@ -46,7 +46,8 @@ const generateVersionedDocs = async (versions) => {
     const configFile = `./version/${version}.yml`
     const newSrc = `./version/src_${version}`
     versionedMkdocs.extra.versions = extraVersions
-    versionedMkdocs.docs_dir = newSrc
+    versionedMkdocs.docs_dir = `./src_${version}`
+    versionedMkdocs.theme.custom_dir = '../theme'
     fs.writeFileSync(`./version/${version}.yml`, YAML.stringify(versionedMkdocs, 7), 'utf8')
     fse.copySync(`./src`, newSrc)
     await git.checkout(returnToBranch)
```
dmitry-timofeev: I'd consider clarifying the intent behind the steps of this code. It is absolutely not clear why it first switches to `master` (returnToBranch) and then builds the docs, and it works (it seems it works because it makes a copy of the sources). The fact there was a bug just confirmed it begs the clarifications of the intent.

package.json
dmitry-timofeev: FYI, there is an interesting dockerfile in autotests, that seems to install every tool on earth for link checking (the job might use some of them): 

- https://github.com/exonum/autotests-exonum-java-binding/blob/master/images/links-verify/Dockerfile 
- http://192.168.221.185:8888/view/EJB/job/exonum-java-binding-links-verify/

versioned.js
slowli: I have no idea tbh; I didn't write the script.

package.json
slowli: I believe the name should be interpreted as "(broken link) checker", not "broken (link checker)" 🙃

versioned.js
dmitry-timofeev: Well, I am not saying "rewrite the whole thing" but "since we've already understood the intent, why don't we clarify it for future mainteiners — possibly us — through some implementation comments/clearer structure/etc so that they don't have to invest some time to uncover it each time it needs to be modified".

slowli:
@dmitry-timofeev Let's fix / improve the versioning script separately; for now, I'll merge this PR as is.

dmitry-timofeev:
Okay

PR #419, opened 2020-04-01
Add action field to launcher configs examples

vitvakatu:

### DIFF
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -1103,7 +1103,8 @@ refactoring. Java artifacts also need the name of the JAR file in the
 `spec: artifact_filename` field of the artifacts directory. The present example
 shows how to add the Java `cryptocurrency-demo` service, and two Rust services —
 the `timestamping` and `time` oracle services. Please note, that Rust services
-are pre-deployed in the genesis block, so we don't use `action` field for them.
+are pre-deployed in the genesis block, so we do not specify `action: deploy` for
+them.
 
 ```yaml
 artifacts:
--- a/src/get-started/java-binding.md
+++ b/src/get-started/java-binding.md
@@ -94,7 +94,7 @@ follow the instructions in [Contribution Guide][how-to-build].
     export PATH="$PATH:$EXONUM_HOME/bin"
     ```
 
-2. Install [JDK 11][jdk].
+2. Install JDK 11.
 
 3. Install [Maven 3][maven-install] which is essential for developing and building
   Java service.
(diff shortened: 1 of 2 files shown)


src/get-started/java-binding.md
```diff
@@ -1102,7 +1102,8 @@ aliases are used in other parts of the configuration for readability and easier
 refactoring. Java artifacts also need the name of the JAR file in the
 `spec: artifact_filename` field of the artifacts directory. The present example
 shows how to add the Java `cryptocurrency-demo` service, and two Rust services —
-the `timestamping` and `time` oracle services.
+the `timestamping` and `time` oracle services. Please note, that Rust services
+are pre-deployed in the genesis block, so we don't use `action` field for them.
```
dmitry-timofeev, line 1106: We do for services, therefore, I'd clarify that we do not _deploy_ them:

... we don't specify `action: deploy` for them.

... we do not deploy their artifacts (= specify `action: deploy`).

src/get-started/java-binding.md
```diff
@@ -1312,7 +1312,7 @@ For using the library just include the dependency in your `pom.xml`:
 [homebrew]: https://github.com/Homebrew/brew#homebrew
 [how-to-build]: https://github.com/exonum/exonum-java-binding/blob/ejb/v0.9.0-rc2/CONTRIBUTING.md#how-to-build
 [java.util.Properties]: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#load(java.io.Reader)
-[jdk]: https://jdk.java.net/
+[jdk]: https://jdk.java.net/java-se-ri/11
```
dmitry-timofeev, line 1315: The reference implementation is not something we shall link:

> These binaries are for reference use only!

> These binaries are provided for use by implementers of the Java SE 11 Platform Specification and are for reference purposes only

I think we shall either drop the link; or link one of the providers, e.g., https://adoptopenjdk.net/?variant=openjdk11&jvmVariant=hotspot (but that is certainly not the only provider)

src/get-started/java-binding.md
```diff
@@ -94,7 +94,7 @@ follow the instructions in [Contribution Guide][how-to-build].
     export PATH="$PATH:$EXONUM_HOME/bin"
     ```
 
-2. Install the [latest JDK][jdk].
+2. Install [JDK 11][jdk].
```
dmitry-timofeev, line 97: Most importantly, tutorial.

src/get-started/java-binding.md
vitvakatu: Yep, it was my concern

PR #423, opened 2020-04-13
Migration tutorial [ECR-4182]

slowli:
This PR adds a migration tutorial.

---

See: https://jira.bf.local/browse/ECR-4182

### DIFF
--- a/src/get-started/migrate-service.md
+++ b/src/get-started/migrate-service.md
@@ -1,0 +1,273 @@
+# Data Migration Tutorial
+
+In this tutorial, we look into transforming a
+[simple cryptocurrency service](create-service.md) into an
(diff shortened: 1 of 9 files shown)


src/get-started/migrate-service.md
```diff
@@ -0,0 +1,273 @@
...
+            .add_script(Version::new(0, 2, 0), migrate_wallets)
+            .select(start_version)
+    }
+}
+```
+
+`LinearMigrations` type allows to choose which script(s) to execute
+given the initial service version.
+The provided migration scripts are associated with a semantic version
+for service data after migration. In our case, we have a single script,
```
aleksuss, line 159:
```suggestion
for the service data after migration. In our case, we have a single script,
```
???

